PR #42, opened 2026-07-07
Projectless daemon: optional project root, no cwd fallback, structured no_project error

lovon-spec:
Phase 1 of the "no project open" story: a daemon launched from a directory with no project marker runs **projectless** instead of adopting cwd as an implicit project. Removes the macos-app `~/projects/intendant-workspace` bandaid at its source.

## Design

- **Detection rule (one definition):** a root is a real project iff it carries a marker — `.git` (directory, or file for git worktrees) or `intendant.toml`. Now owned by `project::root_has_project_marker()`; the 872a64ab boot-scan gate (`file_watcher::root_is_snapshot_worthy`) delegates to it. The *daemon* shape with a markerless launch cwd runs projectless; CLI (headless/MCP) invocations keep cwd-as-project — correct for `intendant "task"` inside a repo.
- **`Option<PathBuf>` through the daemon path:** `main` computes `daemon_project_root` once (from flags + the marker) and threads it through `run_daemon`, `DaemonConfig`, `SessionSupervisorConfig.project_root`, `spawn_mode_web_gateway`, and `start_project_file_watcher`. The gateway routes were already `Option`-ready (`project_root: null`, settings POST 400s "No project root", worktree scan tolerates None) — the daemon just never passed `None` before.
- **Fail-closed consequences of projectless** (absence of a project shrinks scope, never widens it):
  - no file watcher at all (one clear log line), on top of the existing non-project-root scan gate;
  - sandbox write scope = scratch + log dir + `~/.intendant` + explicit **absolute** grants only (`SandboxConfig::projectless`); a relative grant with no project to anchor to is dropped loudly;
  - no project-root `.env` layer at startup (the marker gate also stops the redundant cwd-fallback double-report on CLI runs; the ordinary cwd walk-up still applies);
  - the daemon's base session meta records no project; the git-vitals producer is skipped.
- **Sessions:** every create/resume funnels through one choke point. Explicit `project_root` works exactly as before. A `CreateSession` without one on a projectless daemon fails with structured `SessionEnded { error_kind: "no_project" }` (the unfueled precedent) plus the log-shaped "Session create failed:" line the dashboard already keys on, and an honest failed-session summary. Native resume gains a last-resort fallback to the session's own `session_meta.json` root (reachable only when projectless — rooted daemons behave byte-for-byte as before); external resume: explicit → persisted config → daemon default → structured failure.
- **Dashboard (minimal):** the New Session submit path blocks when the daemon reports `project_root: null` and no project is entered (existing spawn-notice mechanism, "Pick project" action focusing the Project field); the Station launch pane uses the same guard; a `no_project` SessionEnded fails a pending spawn notice with the same action. No wasm changes — presence-web already passes `error_kind` through generically, so `static/wasm-web` is untouched.
- **macos-app:** the synthetic-workspace fabrication and marker walk-up are deleted; cwd is plainly `$HOME`. The workspace dir remains supported as user data — it just stops being auto-created/assumed.

## Evidence

- `cargo check --bins` and `cargo check --test e2e` clean; new unit tests added (projectless sandbox scope parity, marker-gated resolve errors, no-project resolve classes).
- New headless e2e scenario `projectless_daemon_serves_and_requires_an_explicit_session_project` (mock provider, no keys/network/display): boots the daemon from an empty temp cwd, asserts the projectless startup lines + `project_root: null` over HTTP, drives `/ws` for (a) a rooted CreateSession completing a scripted exec round-trip through the real runtime and (b) the structured `no_project` failure for a rootless one — the required cross-platform check runs it on all three platforms.
- Local full battery (`cargo test --bins`, `cargo test --test e2e`, clippy, manual projectless boot check) is running on a heavily loaded box; results will be posted as a PR comment before review sign-off.

## Phase 2 (tracked in the design handoff)

- Dashboard "no project open" empty state (project picker front-and-center) + the Activity-composer path (`StartTask` with no session currently gets only the structured failure, no composer UX).
- Per-session sandbox write scope (`INTENDANT_SANDBOX_WRITE_PATHS` is daemon-global; with `--sandbox` a projectless daemon's sessions can't write their own project root — pre-existing, now more visible).
- An explicitly configured default project for daemons; `--continue` on a projectless daemon; settings persistence without a project root (global config home); terminals defaulting to a session's project; fresh-VPS/service re-verification under projectless.

---
Supersedes #35: GitHub stopped creating check suites for that PR's heads (zero suites for two consecutive pushes; reopen and synchronize both undelivered), so it was recreated to re-mint Actions dispatch. Diagnosis history and the e2e root-cause comments live on #35.

### DIFF
--- a/docs/src/control-plane-and-daemon.md
+++ b/docs/src/control-plane-and-daemon.md
@@ -127,7 +127,11 @@ When `start_new_session` runs, it:
 1. resolves a fresh session log directory (`SessionLog::resolve_path(None)` →
    `~/.intendant/logs/<uuid>/`) and opens the `SessionLog`;
 2. resolves the project root (per-session override or the daemon's default) and
-   loads the `Project`;
+   loads the `Project`. On a **projectless daemon** (see below) there is no
+   default: a `CreateSession` without an explicit `project_root` fails with
+   `SessionEnded { error_kind: "no_project" }` — a structured class the
+   dashboard turns into the project picker — instead of silently rooting the
+   session at the daemon's launch directory;
 3. writes `session_meta.json` and activates the shared active-session handle;
 4. resolves the backend (one-shot `agent` override → configured default →
    internal) and applies the runtime external-agent config;
@@ -179,7 +183,8 @@ by it, so the supervisor picks it up.
 
 ## File Watcher, Snapshots, and Rewind
 
-`file_watcher.rs` is a live filesystem watcher rooted at the project directory.
+`file_watcher.rs` is a live filesystem watcher rooted at the project directory
+(a projectless daemon starts no watcher — there is nothing to watch).
 It works for **all** agent types — internal, Codex, Claude Code —
 because it watches the filesystem directly rather than diffing git, so an
 external CLI's edits show up the same as Intendant's own. It does two jobs:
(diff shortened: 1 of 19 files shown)


lovon-spec:
Recreating on a fresh ref: GitHub is not delivering Actions events for this branch (zero check suites across three heads and three PR events); same commits continue on worktree-projectless-daemon-b.

PR #98, opened 2026-07-08
ci: probe skipped required checks against the merge queue (task #45)

lovon-spec:
One-shot experiment per task #45: the required `test (…)` jobs are guarded with `if: github.head_ref != 'ci-skip-probe'`, so this PR's own pull_request runs mint **created-then-skipped** required check runs while the merge-group gate still executes normally on the fleet.

- Merges cleanly → the queue accepts skipped required checks; a follow-up removes the guard + `.github/PROBE.md`, and the two-job hosted-relevance pattern (doc-only PRs off the fleet) is viable.
- Wedges at Expected/pending → the queue rejects them; this PR gets closed and the branch deleted.

### DIFF
--- a/.github/PROBE.md
+++ b/.github/PROBE.md
@@ -1,0 +1,17 @@
+# CI skipped-check probe (task #45)
+
+Scratch marker for a one-shot experiment: the PR carrying this file
+guards the required `test (…)` jobs with
+`if: github.head_ref != 'ci-skip-probe'`, so its own pull_request runs
+mint **created-then-skipped** required check runs while its merge-group
+gate still executes normally on the fleet.
+
+Question under test: does the merge queue treat a skipped required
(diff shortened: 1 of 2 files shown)


lovon-spec:
Probe answered at an earlier layer than designed: a job-level `if:` on a matrix job skips **before matrix expansion**, minting a single check named literally `test (${{ matrix.os }})` (completed/skipped) — the pinned per-leg required contexts are never created, so this wedges as name-mismatch regardless of whether the queue accepts skipped checks. Conclusion: the two-job hosted-relevance pattern is unusable for matrix jobs with per-leg pinned contexts without de-matrixing into literally-named jobs or a coordinated ruleset rename. The in-job fast path stays. Guard + PROBE.md die with this branch (never merged).

PR #210, opened 2026-07-11
wasm-drift probe: exercise the gate's rebuild path

lovon-spec:
Deliberate probe, held out of the queue until its wasm-drift check reports: a comment-only touch to presence-web takes the gate down its rebuild path for the first time (every run so far green-skipped on relevance). Codegen is unchanged by a comment, so a green run proves BOTH the rebuild plumbing (pinned wasm-pack install on the runner account, external cache, diff) AND cross-host byte-determinism (committed artifacts are macOS-built; the gate rebuilds on Linux). Red means macOS-built artifacts can't serve as the committed canon and the gate needs a Linux-regen or an advisory downgrade before promotion.

### DIFF
--- a/crates/presence-web/src/lib.rs
+++ b/crates/presence-web/src/lib.rs
@@ -1,4 +1,5 @@
-//! Browser-side presence layer in Rust+WASM.
+//! Browser-side presence layer in Rust+WASM. (Comment-only touch: the
+//! wasm-drift gate's rebuild path was first exercised by this line.)
 //!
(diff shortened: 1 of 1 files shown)


lovon-spec:
Probe served its purpose: it proved the rebuild path executes, that wasm output is NOT byte-deterministic across host triples (moving the gate to the Mac leg), and surfaced the cold-cache install race (fixed by rebuild serialization). The registry-path remap (#219) closed the cross-account gap it exposed. Closing unmerged — the comment touch has no lasting value.

PR #298, opened 2026-07-13
security: require trusted anchors for daemon authority

lovon-spec:
## Summary

- Makes hosted Connect strictly discovery and route metadata. A claim creates no daemon principal or grant, hosted provenance is immutable `role:none`, Connect no longer serves the daemon SPA/control client, and both service and daemon reject legacy offer/ICE/close traffic before touching control or IAM state.
- Requires a trusted anchor for daemon authority in the default product: loopback/local presence or an independently reached, fingerprint-verified direct-mTLS surface. `--owner` is retired, legacy hosted-root and browser-key owner grants are revoked by the schema-v2 migration, and there is no same-build risk opt-in. No signed/notarized native remote anchor ships in this alpha.
- Treats rendezvous-controlled fleet WebPKI names as discovery-only at the TLS/HTTP boundary. Protected HTTP, MCP, signaling, and WebSocket traffic is refused on fleet provenance even when the browser presents an enrolled client certificate.
- Tightens direct enrollment and migration: owner bundles are released only after browser-certificate fingerprint transcription, enrollment always serves the exact access certificate rather than fleet SNI, request authentication cannot mint root, and trusted startup migration recognizes only the daemon-generated owner certificate. Windows now supports local setup/recert/list/remove with explicit CurrentUser certificate-import instructions.
- Makes long-lived authority revocable in practice: dashboard, WebSocket, peer, file-transfer, display, input, and clipboard paths revalidate the exact opening grant/identity; buffered input is bounded, ordered, epoch-gated, and safety-releases held keys/buttons on revoke, transport loss, overflow, or injection failure.
- Reframes the twelve words as a single-use claim code, keeps registration and route-link changes daemon-signed, and updates README/headline docs, detailed trust/custody/rendezvous/dashboard chapters, installers, Connect copy, and Access UI to state the shipped limits accurately.

## Root cause

The old promise treated a browser credential as a root anchor even though future JavaScript served by that hosted origin could invoke it. A passkey/PRF gesture protects key extraction but does not make hostile same-origin code trustworthy. Separately, some long-lived streams and queued input could survive changes to the grant that opened them. The fix moves authority to daemon-controlled anchors and binds every retained capability to live local IAM/peer state.

## User and developer impact

- Hosted claiming can link and find a machine, but cannot observe or operate it.
- Remote control requires `intendant access setup` from local console/SSH and an independently verified direct-mTLS route; loopback remains local root.
- The public fleet URL can serve the inert shell/discovery bytes but cannot become a control origin.
- Existing legacy `connect-bootstrap` and `--owner` browser-key authority is revoked and must be re-enrolled through direct mTLS.
- The current unsigned development macOS bundle is not represented as a distribution trust anchor.

## Validation

- `cargo test --bins`: 3,373 passed, 9 ignored
- `cargo test -p intendant-display --lib`: 528 passed, 2 ignored
- `cargo test --test e2e -- --test-threads=1`: 20 passed
- `cargo clippy --workspace --locked -- -D warnings`
- `cargo fmt --all -- --check`
- `mdbook build docs`
- `node scripts/validate-connect-hosted-mvp.cjs`
- `node scripts/connect-transport-e2e.cjs`
- `node scripts/validate-connect-rendezvous.cjs --dashboard-binary target/debug/intendant`
- Access-focused suite: 200 passed; dedicated single-cert enrollment resolver test passed
- Linux fleet `cargo check --bin intendant` passed
- Generated `static/app.html` parity, all changed JS/CJS syntax checks, installer shell syntax, `CLAUDE.md`/`AGENTS.md` parity, and `git diff --check`

Native Windows SSH was unreachable during the local pass, and the installed macOS cross target could not build `ring` without an MSVC SDK. The draft PR's native Windows `cargo check` gate passed on the self-hosted Windows runner.

## Scope note

This closes trust/bootstrap Blocker #1. It does not implement the separate future power-user binary or claim that release-channel immutability is solved. This PR intentionally remains draft and is not queued.


src/bin/caller/connect_rendezvous.rs
```diff
@@ -918,29 +752,30 @@ fn note_register_response(response: &RegisterResponse, base_url: &Url) {
...
             status.claim_binding = None;
-            let (effective_code, effective_url) = match &bootstrap {
-                Some((phrase, url)) => (Some(phrase.clone()), Some(url.clone())),
-                None => (response.claim_code.clone(), response.claim_url.clone()),
-            };
+            // New services retain only the daemon-minted hash. Keep parsing
+            // legacy plaintext fields, but prefer the local code that was
+            // actually signed into this registration.
+            let local_code = peek_route_claim_code();
+            let effective_code = local_code.clone().or_else(|| response.claim_code.clone());
```
lovon-spec: **[P1-compat]** This preference order breaks claiming against an older service: a legacy service ignores `claim_code_hash` and mints its own plaintext code, but the daemon displays the local code the old service never learned, so the presented code can never match its store. Preferring `response.claim_code` when non-null restores mixed-version claiming at equal security — post-refactor a claim code grants nothing, and new services return `claim_code: null`, so the locally-minted code still wins against them.

lovon-spec (commented):
Deep review at 80a94be2 across the IAM core + migration, the connect service, fleet-provenance/TLS, the gateway surface, and input/display/transfer revocation. **The core claims hold and are fail-closed everywhere I checked**: hosted provenance is immutably `role:none` even against deliberately minted grants; legacy `connect-bootstrap` root grants are revoked by a sticky, fail-closed startup migration; fleet provenance is SNI-anchored (registered before cert issuance, never forgotten) with the Host fallback only ever removing authority, enforced ahead of identity resolution on every control lane; the datachannel tunnel can't be opened from fleet provenance, and every lane revalidates the exact opening grant (byte-equal grant/principal/role + hosted_origins stability) per message/event/chunk plus a 250ms tick; daemon registration is Ed25519-signed with monotonic `issued_at`, and every polling endpoint requires the rotating session token; the docs rewrites match the code. This is ready to come out of draft once the items below are addressed.

**1. [P1] Revoked clients keep receiving display video on the dashboard-control lane.** `api_display_webrtc_offer_response` (`src/bin/caller/dashboard_control/api_control.rs:274`) registers the media `WebRtcPeer` with no teardown tied to the grant. On revocation the control driver dies within ~250ms and input/clipboard fail closed, but the media peer is never closed — a hostile client keeps the live video stream until it hangs up. The `/ws` lane (`remove_peer` on session close) and the federated lane (`close_federated_peers_for_sessions`) both tear down display peers on grant loss; this lane is the one gap in "display … paths revalidate the exact opening grant". Suggest closing the media peer from the control-driver teardown / authority tick.

**2. [P3] Test gap:** every fleet-SNI test uses `with_no_client_auth()`; nothing pins that a fleet-SNI connection presenting an *enrolled client cert* is still refused for protected traffic. The ordering guarantees it structurally (the fleet gate precedes cert/grant resolution on both the HTTP and WS paths) — worth one test so the ordering can't silently regress.

**3. [P3] Question:** `access/enrollment.rs` demotes `record_refused_client_key` and `MAX_PENDING` to `#[cfg(test)]` — is the pending-enrollment doorbell now writer-less in production (dead UI), or is refusal recording handled elsewhere after this refactor?

**4. [P3] Inert root grants are still mintable for hosted-origin keys** (verified inert at exercise — `hosted_sessions_are_immutably_denied_while_direct_keys_remain_authorized`). They render as ordinary active grants, though; a mint-time refusal or an Access-UI marking would avoid a misleading "active root" row.

Deliberate behavior changes flagged for ratification, not objection: `/config` now requires `presence.read` + own/fleet origin (narrowly-scoped custom grants lose TURN bootstrap); cleartext non-loopback own-origin is refused, so a `--no-tls --bind <routable>` daemon is loopback-authority-only; a direct-mTLS client sending a fleet `Host:` is 403'd (fail-closed footgun — maybe a docs note); input-queue hard-cap overflow now trips the display's whole browser-input lane until the display session is recreated (correct trade vs stranded keys, but operationally new).

Four more findings are inline on the diff. Remaining test gaps worth a follow-up pass: mid-stream transfer halt on revocation (the per-chunk enforcement itself, not just the flag flip), clipboard revalidation in both directions, the 250ms tick close on the WS/tunnel lanes, and the listener authority-observer reset integration.

src/bin/caller/web_gateway/ws_session.rs
```diff
@@ -1693,16 +1820,22 @@ pub(crate) async fn ws_inbound_task(
...
-                                .unwrap_or_else(|e| e.into_inner());
-                            match authority.get(&display_id) {
-                                Some(entry) => entry.matches_local_ws(&connection_id_inbound),
-                                None => true,
-                            }
-                        };
-                        if !allowed {
+                        let input_authorized = Arc::clone(
+                            local_display_input_authorizers
+                                .entry(display_id)
```
lovon-spec: **[P2]** This `or_insert_with` runs before the authorization/session lookup succeeds, and `display_id` is a client-chosen u32 — a denied-but-connected client can grow this per-connection map one Arc'd closure per distinct id across the u32 space, never evicted until WS close. Same pattern in the `display_offer` arm (~line 1393), where the entry persists even when the offer 404s. Suggest inserting only after the session lookup/authorization succeeds, or bounding the map.

src/bin/caller/web_gateway/listener.rs
```diff
@@ -311,14 +308,79 @@ pub fn spawn_web_gateway(
...
+                            None => {
+                                last_holders.remove(&change.display_id);
+                            }
+                        }
+                        // Every holder transition advances the display input
+                        // epoch and clears its backlog, even when no native
+                        // edge is held yet. Otherwise a queued A event could
+                        // survive a fast A -> B -> A sequence and become true
+                        // under A's identity-only guard again.
+                        if identity_changed {
```
lovon-spec, line 543: **[P2]** Two independent resets fire for one A→B authority transition: `push_guarded`'s inline revision-mismatch reset when B's first event arrives, and this observer's unconditional `reset_browser_input`. If B's events land between the two, the second reset discards B's already-admitted backlog — the new holder's first clicks/keys vanish silently on the fire-and-forget lane. Making this reset conditional on the queue's admitted `authority_revision` predating the change's revision (the queue already stores it) would close the race. Related wart: a re-grant to the *current* holder also bumps the revision — the observer skips same-identity transitions but `push_guarded` can't, so a redundant grant mid-drag force-releases the holder's own held button.

src/bin/caller/dashboard_control/api_control.rs
```diff
@@ -321,7 +317,14 @@ pub(crate) async fn api_display_webrtc_offer_response(
             Some(Arc::clone(&runtime.tcp_peer_registry)),
             runtime.tcp_advertised,
             ice_tx,
-            input_authorized,
+            crate::display::BrowserInputAuthorization::versioned(
+                input_authorized,
+                runtime
+                    .display_authority
+                    .as_ref()
+                    .expect("display offer requires authority bridge")
```
lovon-spec: **[P2]** This `.expect` turns a browser-supplied `display_offer` frame into a daemon panic whenever `display_authority` is `None`. Unreachable in production today (the listener always passes `Some(bridge)`; only tests construct `None`), and the twin in `dispatch.rs` is guarded by `dashboard_display_input_remains_authorized` (`None` ⇒ deny) — but a future embedder shouldn't get a panic from a remote frame. Prefer an error frame.

lovon-spec:
One more small thing found while landing the docs-staleness batch (#311): `tests/e2e/main.rs` is at **21** `#[tokio::test]` cases on current main, and this branch's rewritten getting-started passage says 20 — worth touching up on your next merge of main.

lovon-spec (commented):
**Fix-round re-review at 7efc789c** — delta review of the response commit against the previous review's 8 findings (both halves: IAM/fleet/gateway and display/input/tunnel), verified line-level against the checked-out tree.

## All 8 previous findings are genuinely fixed

Each fix comes with a behavioral test that fails on the pre-fix code (checked):

1. **P1 display media** — media peers are now tracked in `display_peer_sessions` and closed by `remove_dashboard_display_peers` on the control driver's single exit tail (`wire.rs:352`), so revocation kills video within the same ~250 ms window as input/clipboard. Both offer races are closed (pre-registration rollback and the post-offer authority-expired rollback at `api_control.rs:345`).
2. **P1-compat claim code** — a legacy service's echoed `claim_code`/`claim_url` now wins as a coherent pair; new services return null so the local code still wins there.
3. **P2 pre-auth map growth** — both `or_insert_with` arms moved behind authorization, and the global authority maps are covered too (2048-ID storm tests).
4. **P2 handoff double reset** — observer resets are now conditional on `reset_before_authority_revision` (serial-compare), so one reset per transition, and same-holder re-grants no longer reset at all.
5. **P2 `.expect` panic** — now a 503 error frame.
6. **P3 fleet-SNI + client-cert** — test added, with a positive control.
7. **P3 enrollment doorbell** — declared as a staged capability; `writer_available: false` self-reported end to end.
8. **P3 inert hosted root grants** — refused at mint, reactivation, *and* exercise; projected honestly as `inactive_binding`; durable across restarts via `fleet-origin-provenance.json`, fail-closed on incomplete provenance.

Also: dropping the docs' hardcoded e2e count entirely beats bumping it — that one stays fixed.

## Two new findings introduced/exposed by the fix round

1. **[P2] Cross-lane PeerId collision — the new teardown paths can close the wrong lane's peer.** `NEXT_DASHBOARD_DISPLAY_PEER_ID` (`dashboard_control/mod.rs:72`) and the `/ws` lane's `NEXT_PEER_ID` (`web_gateway/mod.rs:85`) are independent `AtomicU64`s that both start at 1, and both lanes register peers into the same `DisplaySession.peers` map. The first control transport and the first `/ws` tab watching the same display both hold peer id 1. The collision predates this PR, but this commit adds removal paths keyed by that id — `remove_dashboard_display_peers` (`api_control.rs:387`, called from the driver exit at `wire.rs:352`) and the post-offer rollback (`api_control.rs:345`) — so an IAM revocation of a hosted control session can now tear down a local `/ws` viewer's live media peer instead of (or in addition to) its own. Fix direction: one shared allocator for both lanes, or lane-namespaced id ranges. This should be fixed or explicitly accepted before landing.

2. **[P3] `get` → `get_any` widening: private user views are now reachable by scoped, non-owner principals.** The control lane now resolves displays via `get_any` (`api_control.rs:443`, `:639`; `dispatch.rs:2201`), which includes `agent_visible=false` user-session views. The comments justify it as "this is an owner dashboard lane" — but the dashboard-control lane is exactly the lane a *scoped delegate* (non-owner client key with display permissions) drives, so the premise doesn't hold post-trust-model. Either gate private-view resolution on the actor (owner/root → `get_any`, scoped → filtered `get`), or state explicitly that display-scoped delegates are intended to see private user views. Either way, `ws_session.rs:1430`'s claim that private user views stream on `/ws` "and only here" is now false and should be updated.

## Smaller residue (take or leave)

- [P3] `DEFAULT_HOSTED_ORIGIN` is compiled in, but the default fleet *zone* isn't — a Connect-disabled daemon classifies another daemon's fleet-route origin as `direct` at mint and exercise. Inert today (no shipped transport presents client-key proofs there); a compiled default zone would be the symmetric fix.
- [P4] Provenance-incomplete refusals reuse the "fleet origins" reason string — a distinct reason plus a repair hint would save operators a confusing dead end.
- [P4] `display_peer_sessions` never prunes cancelled sessions' Arcs (grows for the daemon's lifetime under session churn).
- [P4] `apply_dashboard_grant_for_existing_display` maps `try_read` contention to 404; the `/ws` lane awaits the lock instead — 503 (or awaiting) fits better.
- [P4] `client_key.rs`'s module doc still says the verify path has "no production ingress," but `resolve_peer_offer_attribution` calls it.
- [P4] `spawn_test_gateway_with_tls_acceptor` defaults to the runner's real cert dir (hermeticity).
- [P4] Control-lane `display_id_param` still `unwrap_or(0)` where `/ws` rejects malformed ids strictly.
- [P4] `remove_peer`'s "sole caller" doc comment is stale now that three paths call it; `access_enrollment_decide_response_value` consumes the queue entry before the upsert guard runs.

Coverage gaps worth one pass (non-blocking): nothing drives the control driver's exit end-to-end to pin the `wire.rs:352` teardown; the post-offer 403 rollback is untested; the provenance-refresh path is only exercised by calling the restore fn directly; the legacy code-with-null-URL claim print is unpinned.

**Verdict: with finding 1 fixed (or explicitly accepted in a comment) and finding 2's intent stated, this is ready to un-draft from my side.**

lovon-spec:
**Heads-up: the branch is now `CONFLICTING`/`DIRTY` against main** — 17 commits have landed since the last main merge (`40c4be74`), mostly the composer program (#315/#316) and #314. While a PR is DIRTY it has no merge ref, so GitHub mints no check suites and even armed auto-merge would park forever; the next push needs a merge of main into the branch (merge, not rebase).

A merge simulation (`git merge-tree --write-tree`) shows exactly two conflicting files:

- `static/app.html` — both sides regenerated the artifact. Per repo policy, don't hand-reconcile it: resolve the `static/app/` fragments, run `cargo run -p app-html-assembler`, and commit the regenerated file.
- `docs/src/configuration.md` — a genuine content conflict, docs-only.

Everything else, including the Rust files both sides touched (`api_control.rs`, `http_dispatch.rs`, `main.rs`, `web_gateway/mod.rs`), auto-merges cleanly.

lovon-spec (commented):
**Round-3 verification at `631631cc` (+ the follow-up main merge `fb940334`): GO — ready to un-draft from the review side.**

Everything committed in the previous round is delivered, each with regression tests verified to fail on the pre-fix code:

- **PeerId collision** — fixed via the new `display_peer_ids.rs` (2-bit lane namespaces, one shared payload counter for the local lanes, fail-closed `None` on exhaustion). Every production mint site routes through it (control `mod.rs:1636`, `/ws` `listener.rs:1686`, federated `routes_peers.rs:1968`), and `display_peer_namespaces_coexist_and_teardown_independently` asserts the exact old failure (three lanes on one session, control teardown removes only its own). Re-keying the federated lane on (connection, session) also **closes a pre-existing cross-connection teardown vector** — the old session-id-only hash let one federation connection tear down another's peer; `same_session_id_on_two_federation_connections_tears_down_independently` pins it. The connection id being a server-assigned UUID makes cross-connection collision infeasible. Nice catch going wider than the finding.
- **Private views** — root/owner-only on both lanes via `has_owner_dashboard_authority()` (TrustedLocal, or root with live opening authority; peers never) at every resolve site; non-owner display listings redact private capture state; deny tests on both lanes fail on the old `get_any` code; the residual `get_any` calls are lifecycle-internal only (safety-release, teardown). No mid-stream public→private downgrade leak either — narrowing evicts peers via a fresh session.
- **Dashboard-action authorization** — two layers, each declared once: per-action operation floors (`control_msg_access_decision`) so no action is reachable at Message level anymore, plus the owner-dashboard-only subset (`control_msg_requires_owner_dashboard`) enforced on the tunnel, on `/ws` ahead of the peer fast-path, and on MCP (`is_enrolled_root_mtls_user_client` narrows the owner anchor without widening `is_owner_surface`). Leaving revoke/stop de-escalations reachable at their operation floor is the right call. The allowlist itself didn't change, so the SPA parity pin holds.
- **recording.rs** — the +510 is justified, not creep: recordings land on the agent-readable session tree, so private/absent displays are now refused before ffmpeg spawns (`get` not `get_any`, x11grab fallback removed), the frame bridge re-binds to the exact session Arc each tick (display-id-reuse TOCTOU), and recording lifecycle moved to the lossless intent lane so controls can't be flood-dropped.
- **The four promised P4s** — all delivered: hermetic TLS test helper (injected tempdir, abort-on-drop handle), strict display-id parsing across signaling/authority/dispatch, the `client_key.rs` module-doc correction, and `display_peer_sessions` → `Vec<Weak>` with retain-pruning plus a shutdown-aware reaper (`display_peer_tracking_does_not_retain_finished_sessions`).
- Docs are accurate without overclaiming (web-dashboard.md is notably honest about lifecycle-metadata leakage, matching the event-line filter), deferred items are properly untouched, no history rewrite, and both conflict resolutions are clean — the green app-html gate on the merged head confirms the regenerated artifact.

Independently verified: full unit suite green locally (3437/3437, clippy clean) and all required CI checks pass on the current head.

Take-or-leave residue (P4, none blocking — fine as follow-ups):

1. `set_diagnostics_visual_marker` is the one display-id parse the hardening sweep missed — still `unwrap_or(0)` on both lanes (`api_control.rs:1284`, `ws_session.rs:2189`). Owner-gated, so worst case is a malformed owner request annotating display 0, but the strict-parse treatment would complete the sweep.
2. `DisplayReady` (machine-emitted) now rides the unbounded lossless intent lane (`event.rs:2124`) — rapid display flap under a wedged consumer is unbounded growth, the lane's documented failure mode; worth a bound or coalescing.
3. HTTP `/api/displays` derives `include_private` as `role_id == "role:root"` (`http_dispatch.rs:1045`) instead of the shared `has_owner_dashboard_authority()` helper — no hosted exclusion or liveness check on that path; metadata-only exposure.
4. The marker's owner-check in `access_gates.rs:354` is a string-literal mirror of `control_msg_requires_owner_dashboard` rather than derived (justified by the raw-frame wire shape, but worth a comment linking the two).
5. `send_display_authority_event` (`wire.rs:448`) drops authority updates for scoped clients on lock contention — fail-closed and re-requestable, purely an observability nit.

Nothing left blocking un-draft + queue from my side.

PR #304, opened 2026-07-14
Display view: fix activity-feed overlap, clamp overlay chips, cursor presence halo

lovon-spec:
Fixes and polish for the Display tab's live workspace, from a 2026-07-13 user report.

## Activity feed rows overlapped when content wrapped (bug)
`.ui2-live-activity-row` declared `min-height: 31px`, which replaces flexbox's `min-height: auto` shrink floor — once entries exceeded the feed's `max-height`, flex squashed **every** row to exactly 31px and wrapped content (percent-encoded `type()` blobs, two-line CU action rows) painted over neighboring rows and timestamps. Fixed by making rows non-shrinking (`flex: 0 0 auto`): rows size to content, the list scrolls.

Reproduced before the fix with a CDP probe against a live mock daemon: 17 rows, all exactly 31px tall, 11 with overflowing content. After: zero overflows, zero box overlaps.

## Long raw payloads clamp to a readable preview
Raw CU calls can run to thousands of characters. Rows over 160 chars now collapse to a readable preview — URL-decoded prefix when the text is percent-encoded, plus a total char count (`… (N chars)`) — with the full literal call behind a real disclosure `<button>` (keyboard + screen-reader accessible). Expanding always shows the literal raw call; only the preview decodes.

## Shared-view focus notes clamp into the stage
The focus-region note was a CSS `::after` anchored below-left of the box with no clamping — under the stage's `overflow: hidden` it clipped into invisibility at the bottom edge and cut off at the right edge (both captured in before shots). It is now a real child element the renderer measures and clamps: below the box when that fits, flipped above when it doesn't, never past the canvas edges, and sized to its own text (`width: max-content`) instead of the focus box's sliver width.

## Cursor presence halo
On a busy screen the small arrow cursor is easy to lose. A soft breathing halo ring (iris, ui-v2 tokens) now centers on the agent pointer hotspot, riding the existing cu_action cursor engine — same lerped motion, same idle fade, no new wire data. It hides entirely while the dashboard user holds input authority (no double cursor in Take-control mode) and respects reduced motion.

## Rail polish (in-scope surfaces only)
Activity row hover wash, feed height 190 → 240px, `overflow-wrap: anywhere` hardening on raw payload lines.

## Verification
Driven end to end against a keyless mock daemon (`PROVIDER=mock` + `INTENDANT_MOCK_DISPLAY=synthetic`) with a headless-Chromium CDP driver: synthetic display granted through the real share flow, CU activity injected through the real `cu_action` handler, focus chips through `intendant ctl shared focus` (the real MCP → shared_view path). Before/after screenshots (dark + light) and numeric probes verified by eye; evidence paths listed below (not committed).

- feed overlap before/after: `scratchpad/shots/before-feed-rail-dark.png`, `after-feed-rail-{dark,light}.png`
- chip clamping before/after: `before-chip-bottom-{left,right}-dark.png`, `after-chip-*-{dark,light}.png`
- halo/ripple: `after-overlay-{ripple,cursor}-{dark,light}.png`, take-control no-double-cursor shot
- local battery: `cargo fmt --check`, `cargo nextest run --bins`, `cargo clippy`, static-script parse check


lovon-spec:
Verification detail (headless-Chromium CDP driver against a keyless mock daemon with the synthetic display backend; real share-grant flow, real cu_action handler, real ctl -> MCP -> shared_view path):

**Feed overlap probe** (17 entries incl. two 478-char percent-encoded type() payloads):
- before: `{"rows":17,"contentOverflow":11,"boxOverlap":0,"heights":[31 x17]}` — every row squashed to exactly 31px, 11 rows painting over neighbors
- after: `{"rows":17,"contentOverflow":0,"boxOverlap":0,"heights":[31,31,31,31,31,50,50,117,50,84,31,50,50,50,50,117,49]}` — rows content-sized, list scrolls

**Disclosure toggle**: preview 85 chars → expanded 478 chars, `aria-expanded` tracks, zero overlaps while expanded.

**Cursor halo**: settled-center probe `dx=0.00 dy=0.01` px against the engine's target point; take-control probe `{userDriving:true, cursorOpacity:0.26, haloOpacity:0}` — halo fully hidden while the human drives.

**Focus chips** at hostile corners (bottom-left, bottom-right, top-right): all four clipped-edge booleans false in every case; bottom cases flip above the box, right cases clamp inside the stage. Verified visually in dark + light.

Local battery: `cargo fmt --check` OK, `cargo nextest run --bins` 3251/3251 passed (9 skipped), `cargo clippy` clean, static-script parse PASS.

PR #332, opened 2026-07-15
external: deliver dashboard image attachments to Claude Code natively

lovon-spec:
## What broke

Images attached from the dashboard to a supervised Claude Code session were silently discarded. Resolution worked (`AgentAttachment::Image` with base64 + path; the session log even said "sent with N attachment(s)"), but the `ExternalAgent::send_message_with_images` **default** forwards text only (`let _ = images`), and `ClaudeCodeAgent` never overrode it — its stdin writer hardcoded a single `text` content block. Never-implemented rather than a regression: the dropping default and Codex's `localImage` override landed together, and the CC backend was born text-only the next day. Non-image files were unaffected (staged-path prelude).

## The fix

- `CcContentBlock` becomes a tagged enum (`Text` | `Image` with an Anthropic-style base64 `source`); `write_user_message` delegates to a block-list writer. CC's stream-json stdin takes the Anthropic Messages content-block shapes — parse acceptance probed live on 2.1.2xx with the adapter's exact spawn flags.
- `send_message` now delegates to `send_message_with_images` (the Codex shape), so the goal-notice/bootstrap preludes and turn accounting stay in one place and image blocks ride the same user message.
- New `ExternalAgent::supports_image_input` capability (default `false`; `true` for Claude Code and Codex). All **three** `external_mode` send sites (main, side, subagent) log a visible session warning when a message carries images the backend cannot deliver — this failure class can never be silent again.

## Validation

- `cargo check`, `cargo clippy`, `cargo fmt --check` clean; external_agent unit suite 373/373 (new serialization test pins the exact image-block wire shape; a mod.rs test pins the fail-closed capability default).
- Live end-to-end vision probe (real `claude` CLI, image block on stdin, model names the image color) — running post-usage-window; PR flips from draft once it passes.

### DIFF
--- a/src/bin/caller/external_agent/claude_code.rs
+++ b/src/bin/caller/external_agent/claude_code.rs
@@ -12,9 +12,9 @@ use tokio::sync::{mpsc, Mutex};
 use crate::error::CallerError;
 
 use super::{
-    normalize_plan_status, AgentConfig, AgentEvent, AgentThread, AgentUsageSnapshot,
-    ApprovalCategory, ApprovalDecision, ExternalAgent, GoalActionOutcome, GoalEngine,
-    SubAgentState, ToolCompletionStatus,
+    normalize_plan_status, AgentConfig, AgentEvent, AgentImageAttachment, AgentThread,
+    AgentUsageSnapshot, ApprovalCategory, ApprovalDecision, ExternalAgent, GoalActionOutcome,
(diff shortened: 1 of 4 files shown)


lovon-spec:
Live vision probe passed post-limit-window: image block via stream-json stdin to the real CLI (haiku, adapter's exact spawn flags) → result `success | red`. Wire shape confirmed end to end.

PR #337, opened 2026-07-15
external-agent: incremental Codex trace/context caches + managed-context I/O fixes

lovon-spec:
Efficiency pass over the external-agent supervision + managed-context trace subsystems (audit scopes 19 + 06, verified against current main). The dominant theme: the managed-Codex context-trace layer re-read its whole archive on every access; it now maintains incremental state and reads only what changed.

## Findings implemented

| Finding | Site | Fix | Complexity |
|---|---|---|---|
| `context_snapshot()` resolved every historical request's chain to return only the newest (~GB reads/turn observed) | `external_agent/codex/context_trace.rs` (`read_latest_codex_context_payload`), `codex/mod.rs` (`read_context_snapshot`) | `CodexTraceIndexCache::latest_snapshot`: snapshot only the max-order request; chain resolution cached and extended (chains are append-only) | O(R²) file reads per call → O(new hops); unchanged-latest preflights re-read 1 payload file |
| Trace index re-read/parsed from byte 0 on every fingerprint change | `codex/mod.rs` (`context_snapshots`), `context_trace.rs` (`read_codex_trace_index`) | per-file byte/line cursors; stat-guarded tail reads fold only appended lines; shrunk/vanished file → full rebuild fallback | O(total lines) per change → O(appended lines); cumulative O(R²) → O(R) |
| Exact-archive fingerprint re-enumerated every session dir under the logs root at 1 Hz | `context_trace.rs` (`collect_codex_trace_bundle_dirs`) | cross-session aux bundle dirs cached behind the logs root's mtime; primary root still enumerated fresh | thousands of read_dir+stat/sec → 1 stat + one small read_dir/sec steady-state |
| `resolved_input.clone()` of the whole conversation only to take `.len()` | `context_trace.rs` (async resolver + sync twin) | count before moving the array into the payload | one full conversation deep-clone per request eliminated (sync twin; async path now clones only into the chain cache, which it repays in avoided rereads) |
| Request sort key allocated 3 Strings per comparison | `context_trace.rs` (`codex_request_sort_key`) | non-allocating comparator `codex_request_order_cmp` | O(R log R) allocations per sort → 0 |
| Every stdout line re-allocated + parsed Value round-tripped through serde | `codex/reader.rs` reader loop | borrow the trimmed line; `decode_jsonrpc_message` moves the 5 envelope fields out of the object (parity-tested against the serde derive) | second pass O(tree) → O(fields); one memcpy per line eliminated; two-pass protocol-watch design untouched |
| `terminal_turns_observed` grew unboundedly for the app-server's life | `codex/reader.rs` | insertion-ordered set capped at 256 keys (oldest evicted); dedup only needs recent terminal events | unbounded → O(1) memory |
| CC tool-result text cloned in full, retained only for a 200-char error snippet | `external_agent/claude_code.rs` (`tool_result_events`) | build the failure snippet first, move the text into the event | one full result-text clone per tool completion eliminated |
| Wrapper-index `upsert`/`record_rollout_path` cloned the whole index per call | `external_wrapper_index.rs` | read-only dirty decision against the cached index (shared predicates); clone only when writing | per-row clone on every session-list pass → clone only on actual change |
| `record_changed_file` did one fsync'd full-ledger RMW per file per branch | `fission_lifecycle.rs` watcher loop + `fission_ledger.rs` (`update_branches_changed_files`) | burst-drain FileChanged events, stage per branch, flush one RMW per log dir; skip the write when nothing new merged | save storm × branches fsync'd rewrites → 1 per burst; stream-order/terminal semantics preserved |
| `wait_for_branch_terminal` re-parsed the whole ledger at 1 Hz | `fission_lifecycle.rs` | (len, mtime-nanos) fingerprint guard (writes land via atomic rename); re-evaluate cached snapshot when unchanged | ~1 MB parse/sec per waiter → 1 stat/sec |
| `get_status` re-parsed the entire session.jsonl per poll (8–17 MB observed) | `lineage_ledger.rs` (`read_lineage_ledger`) | source-independent facts fold split from the derive; per-path cache with byte cursor: stat-only when unchanged, fold appended lines on growth, full refold on shrink | O(file) per poll → O(appended); benefits mcp get_status and the rewind path without touching those callers |

## Correctness notes

- Tail reads consume complete lines only, plus an unterminated final line iff it parses as a complete JSON value (no strict prefix of a JSON object parses) — matching what the historical full re-scan accepted, including writers that never terminate the final line.
- Rewrite blind spot: a same-length in-place rewrite is undetectable by (len, mtime) — the same blind spot the existing fingerprint gate and wrapper-index cache have always had. Shrinks rebuild from scratch.
- Chain-cache invalidation: keyed by (trace root, thread); cleared on truncation rebuilds and re-initialize; a chain that never reaches the cached call key resolves from scratch (rewrites/forks).
- Aux-dir cache staleness window: a new bundle appearing inside another session's existing trace dir is invisible until the logs root's mtime moves — only reachable with two live sessions tracing the same thread, which is unsupported today.
- Not regressed: protocol-watch stays two-pass with passive status parsing on demand; the observe drain and ambient-event separation are untouched.

## Deliberately not done

- Full fission-ledger write debounce (in-memory document + periodic flush): narrows the durability window and needs a crash-consistency argument — the batch fix removes the dominant multiplier without changing when data is durable.
- Findings in `managed_context_ops/**`, `mcp/**`, `web_gateway/**` (rewind multi-scan, anchor-catalog post-pass, edit-branch resolution sweep, latest-entry reverse iteration, skinny `list_records` — audit 06-2/5/6/7/9): outside this change's file scope; `list_records` additionally needs its callers changed.

## Validation

- `cargo fmt --check` clean
- `cargo clippy` clean (touched files also clean under `--tests`)
- `cargo test --bins`: 3334 + 69 + 83 passed, 0 failed
- `cargo test --test e2e`: 21 passed, 0 failed
- New tests: incremental fold/truncation-rebuild/partial-line for the trace cache; chain extension proven by deleting historical payload files before extending; JSON-RPC decode parity vs the serde derive; bounded terminal-key eviction; batched changed-file ledger write; lineage append/truncate/partial-line cache behavior

### DIFF
--- a/src/bin/caller/external_agent/claude_code.rs
+++ b/src/bin/caller/external_agent/claude_code.rs
@@ -1592,25 +1592,28 @@ impl CcReader {
             return;
         }
 
+        // Build the failure snippet before the delta event so the (possibly
+        // large) result text can move into the event instead of being cloned
+        // wholesale just to keep a 200-char error excerpt alive.
+        let failure_message = is_error.then(|| {
+            let message: String = content_text.chars().take(200).collect();
+            if message.trim().is_empty() {
+                "tool error".to_string()
+            } else {
+                message
+            }
+        });
         if !content_text.is_empty() {
             out.events.push(AgentEvent::ToolOutputDelta {
                 item_id: tool_id.clone(),
-                text: content_text.clone(),
+                text: content_text,
             });
         }
 
         self.open_tools.remove(&tool_id);
-        let status = if is_error {
-            let message: String = content_text.chars().take(200).collect();
-            ToolCompletionStatus::Failed {
-                message: if message.trim().is_empty() {
-                    "tool error".into()
-                } else {
-                    message
-                },
-            }
-        } else {
-            ToolCompletionStatus::Success
+        let status = match failure_message {
+            Some(message) => ToolCompletionStatus::Failed { message },
+            None => ToolCompletionStatus::Success,
         };
         out.events.push(AgentEvent::ToolCompleted {
             item_id: tool_id,
--- a/src/bin/caller/external_agent/codex/context_trace.rs
+++ b/src/bin/caller/external_agent/codex/context_trace.rs
@@ -30,6 +30,7 @@ pub(crate) struct CodexResponsePayloadRef {
     pub(crate) response_id: String,
 }
 
+#[derive(Default)]
 pub(crate) struct CodexTraceIndex {
     pub(crate) requests: Vec<CodexRequestPayloadRef>,
     pub(crate) requests_by_call: HashMap<String, CodexRequestPayloadRef>,
@@ -51,24 +52,29 @@ pub(crate) fn codex_context_snapshot_not_ready(err: &CallerError) -> bool {
     )
 }
 
+/// One-shot form of [`CodexTraceIndexCache::latest_snapshot`] for callers
+/// without a live cache (tests, ad-hoc reads). Builds the index fresh and
+/// resolves ONLY the newest request — never every historical request.
 pub(crate) async fn read_latest_codex_context_payload(
     root: &Path,
     thread_id: Option<&str>,
 ) -> Result<CodexRequestPayloadSnapshot, CallerError> {
-    let snapshots = read_codex_context_payloads(root, thread_id).await?;
-    snapshots.into_iter().last().ok_or_else(|| {
-        CallerError::ExternalAgent(format!(
-            "no Codex inference request payload found in {}",
-            root.display()
-        ))
-    })
+    let mut cache = CodexTraceIndexCache::default();
+    cache.latest_snapshot(root, thread_id).await
 }
 
+/// One-shot read of every request snapshot in order — test surface for the
+/// full-archive behavior the cache exposes through
+/// [`CodexTraceIndexCache::snapshots_excluding`].
+#[cfg(test)]
 pub(crate) async fn read_codex_context_payloads(
     root: &Path,
     thread_id: Option<&str>,
 ) -> Result<Vec<CodexRequestPayloadSnapshot>, CallerError> {
-    read_codex_context_payloads_excluding(root, thread_id, &HashSet::new()).await
+    let mut cache = CodexTraceIndexCache::default();
+    cache
+        .snapshots_excluding(root, thread_id, &HashSet::new())
+        .await
 }
 
 pub(crate) fn context_snapshots_from_trace_archive(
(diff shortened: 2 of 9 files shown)


lovon-spec:
Review round applied in 52a35dcc + 6122bf4f (battery green: fmt/clippy clean, 3341+69+83 unit, 21 e2e):

1. Cursor rewrite detection: cursors now record mtime + a 32-byte pre-cursor window; shrink / same-length-mtime-change / window mismatch all rebuild. Reads clamped to the stat snapshot. Doc comment rewritten to state the actual contract. Tests: same-length rewrite, regrown-past-cursor rewrite.
2. Aux sweep: per-root records probed with one stat per collect, re-enumerating only roots whose mtime moved — a bundle created inside an existing session dir after the sweep is now discovered. Test included.
3. Chain cache: entries with an unresolved tail are served only while the tail is still unresolvable; both hit sites re-walk once it becomes resolvable. Test: missing response event arrives later, next snapshot includes the full chain.
4. Lineage grown-arm: strictly greater-than; equal-length + mtime change re-folds. Test included.
5. Replay-vs-eviction: refuting the premature-completion reachability — the known turnId-less replay re-emits the thread's LATEST final answer (keys a couple of turns deep), and the set was never a cross-restart guarantee (reader-task state, always empty after respawn — the pre-eviction code never suppressed ancient replays either; turnId-carrying replays are dropped by the stale-turn guard independently). Cap raised 256 -> 4096 anyway as free insurance; safety analysis documented at the constant and pinned by a 1000-turn churn test.
6. (Optional) Line-index drift: fixed via mid-line cursor state — the deferred newline after a consumed unterminated tail is swallowed without counting a line; equal-timestamp tie-breaks can no longer flip between incremental and rebuilt views. Comparator comment notes the Path::cmp tie-break nuance.

lovon-spec:
Round 2 applied in 9167efbb + 9818be5b (battery green: fmt/clippy clean, 3344+69+83 unit, 21 e2e):

1. Head-window hash: cursors carry a running FNV hash of the first 4 KiB of consumed content, verified alongside the suffix window on every append read (one extra small read). Residual documented as length+mtime+first-4KiB+last-32-bytes. Test: equal-length regrow with byte-identical suffix window (same timestamp) is detected by the head hash alone — the test asserts the suffix-window premise explicitly.
2. Aux probe cadence: per-root stats throttled to a 5 s monotonic interval (AUX_ROOT_PROBE_INTERVAL), probing immediately after a logs-root re-sweep; decision helper unit-tested with synthetic instants; discovery test forces the probe due rather than sleeping.
3. Lineage append-arm hole: cache entries record the last observed (stat_len, mtime); growth beyond the recorded length appends, any mtime change without growth re-folds — including while a partial tail is pending. Test: partial-tail state + same-total-length rewrite with different content and mtime.
4. Optional (a) done: resolved chains record the index generation; an unchanged index serves even a cycle-truncated chain (nothing new could resolve it), any index change falls back to the resolvable-tail check. Optional (b) done: bounded-eviction test crosses the cap with two-key turns.
5. Eviction item: stays refuted per the delta adjudication; the constant's comment now names the per-thread latest-final-answer map as the future structure if a deeper replay source ever appears. CRLF drift remains a documented residual at the comparator.

PR #343, opened 2026-07-16
mcp: stop re-collecting, re-reading, and re-serializing on the status/log hot paths

lovon-spec:
Implements the S/M findings from the MCP-scope efficiency audit (status/log read paths re-doing filesystem + process-table work per call under the state write lock). Each finding verified against current source first; only one 3-line commit had touched these files since the audit base, and every finding reproduced.

## Finding → fix

| # | Finding (audit) | Fix | Status |
|---|---|---|---|
| 1 | `get_status` runs the controller-loop collection (ps spawn + all-run-dirs scan + wrapper/meta reads) 2-3x per call, under the `McpAppState` write lock; same per-call collection in `target_session_phase`/`target_session_source`/control handlers | Split the collection into a state-free raw sample (`collect_controller_loop_raw_status`) and a state-dependent finish (enrich, latest-status fold, stale-marker clear, JSON). Raw sample cached in-state for 1s (interior mutability); enrichment always re-runs against live state so folded phases are never stale on a hit. `get_status` pre-warms the cache **outside** the state locks. Marker-mutating tools invalidate; the control-command twins re-seed with their fresh post-mutation collection. | Done |
| 2 | Session-scoped `get_status` re-reads + re-folds the whole `session.jsonl` per call (O(N^2) cumulative for a polled growing log, under the write lock) | Per-log byte cursor in `McpAppState`; hydration folds only the appended tail (the applier is a pure last-write-wins fold — verified: no log-ring pushes — so tail-only replay reaches the same state, and it no longer regresses live-observed phases to stale log state). Cursor validated against the live file; never advances past an unterminated final line. | Done |
| 3 | `get_logs` re-reads the whole log per poll; its dir lookup is an unmemoized duplicate of the session-log store scan | Byte/line cursor per file (bounded map): a poller paging with `since_id = last returned id` skips the consumed prefix; filtered pollers whose `since_id` lags fall back to the historical full scan. `find_session_log_dir_in_home` now derives from the canonical memoized `SessionLog::find_session_by_id_in_home` (duplicate scan deleted). | Done |
| 4 | `codex_app_server_process_tree_active` eagerly computes a full ps snapshot that the live-root fast path never reads; invoked per matched wrapper per collection | Descendants passed as a lazy `once_with(..).flatten()` iterator through the existing tested `_with_root` seam — the ps spawn now happens only when the root is already dead. Sharing one parent-pairs snapshot across the collection needs `intendant-platform` API changes (outside this lane's file fence); with the root-alive fast path + the finding-1 TTL cache the residue is at most one spawn per dead-root wrapper per second. | Done (snapshot sharing out of fence) |
| 5 | Per-event MCP resource notifications: no subscription tracking, no coalescing, sent inline in the fold loop | `subscribe`/`unsubscribe` record URIs; the listener only notifies subscribed ones (MCP resource notifications are subscription-scoped) and coalesces dirty URIs behind a 100 ms debounce (dirty set + deadline flush in the select loop; flushed on bus close). Also records **both** the event's resource and a deferred control command's — the old single slot dropped the first. | Done |
| 6 | `tools/list` rebuilds and ref-inlines all tool schemas per call (router + manual macro re-runs) | Router definitions and the ~20 manual HTTP definitions built once per process (`OnceLock`); per-request work is the name-keyed profile/managed-context filter plus a clone of the survivors. One unfiltered base + per-call filter (instead of per-(managed, profile) caches) keeps the cache bounded against client-supplied profile strings. Gating is name-keyed, so output is unchanged. | Done |
| 7 | Session-source fallback re-reads + identity-scans the whole log per status poll when the source is not in memory | Memoized: resolved sources into `session_sources`, known-non-external ids into a dedicated set; `NotFound` stays uncached so a log that materializes later still resolves. Tail-only identity scanning skipped — `scan_session_log` lives in `session_identity` (out of fence); the memo removes the per-poll repetition. | Done |
| 8 | `take_screenshot` / `execute_cu_actions` clone the full 1-5 MB base64 PNG for the response | Both sites move the owned value (result taken by value; `screenshot.take()` mirroring `last_screenshot()`'s selection). | Done |
| 9 | Unbounded per-session maps in `McpAppState` | `note_session_ended` prunes the ended session's related-id component once the status map outgrows a soft cap (1024). Under the cap, ended sessions stay resident so post-end status/log queries answer from memory; over it, the next query rebuilds from the persisted log via hydration (correct, one full replay slower). Hydration cursors are cleared on prune so that rebuild is a full replay. | Done |

## Deliberate deviations / skips

- **Finding 1's "guard promote behind `controller_loop_dir_has_observable_state`"**: not applied. Promote's discovery legitimately includes codex processes found via the process tree + wrapper index even when the loop dir has no observable state; the stale-check's guard is only correct because its positive result depends on dir markers. The TTL cache removes the cost motive for the guard.
- **"Other findings" (unranked, non-perf)**: `approve/deny/skip ignore the caller's id` is a correctness/race item, left for its own change. `get_logs cursor spaces incoherent across paths` is a wire-visible design change — the new byte cursor deliberately honors the existing persisted line-id space. The `single resource_changed slot` is fixed at the listener level; `handle_control_command_mcp` still returns one URI per command (a command mutating two resources reports the last) — follow-up material.
- **Lineage-lane dependency**: the per-call ledger fold in `get_status` (`status_ledger_candidate_dirs` + merged lineage/fission reads) belongs to the lineage/external lane; its fold/derive split + facts cache is not on this branch's base, so those sub-items are skipped per the lane split.

## Trust surface

The /mcp IAM gate is untouched: no change to `call_tool_by_name_as_caller`, tool gating semantics, or any authorization ordering. Finding 6 caches only the *definitions*; the name-keyed profile filter still runs per request. Subscription recording only *narrows* what the already-connected stdio peer receives.

## Tests

New: cursor pagination with stable ids, limit-break resumability, replaced-file invalidation (no torn phantom lines), unterminated-tail re-serve, hydration tail-only replay (live phase not regressed), ended-session prune under/over cap.

Battery (all `CARGO_PROFILE_DEV_DEBUG=0`, re-run per review round; latest at head): `cargo fmt --check` clean; `cargo clippy` (workspace) clean; `cargo nextest run --bins` 3493 passed / 0 failed; `cargo test --test e2e` 21 passed / 0 failed.

## Follow-ups (identified in review, deliberately out of scope)

- **Emission-stamped ordering token on lifecycle events.** Hydration replay applies log rows ungated and folds to the log's final state — permanently correct because the log lane is a lossless, order-preserving serialization of the same event stream the live listener folds, but a replay can transiently land a few events behind live (the log-sink consumer's queue lag; self-heals on the next appended rows or live event — contract documented at the hydrate fold site). Eliminating even that transient window requires a monotonic token stamped at emission and carried on both lanes (lifecycle `AppEvent` variants currently carry none; row `ts_ms` is minted at write time downstream of the fork) — a cross-lane change through `event.rs`, the session-log writer/replay, and every lifecycle emission site. Deferred.
- **`ControlMsg::Quit` does not close the stdio MCP service.** `request_quit` sets `should_quit`, but nothing in the MCP shape consumes it for shutdown — `run_mcp_server` returns only when the transport ends. Pre-existing (TUI-era flag); worth its own small change if Quit is meant to terminate the process.
- **Stamp `session_id` on persisted `round_complete` rows (source fidelity).** The writer persists `round_complete` without a session id, so replay reconstructs `RoundComplete { session_id: None }` and every consumer must supply attribution context (the MCP reducer now attributes such rows to the hydration target). Stamping the id at the writer — like every other lifecycle row — fixes it at the source and lets replay reconstruct fully-attributed events; touches `session_log/**`, which is inside PR #344's in-flight diff. Deferred.

### DIFF
--- a/src/bin/caller/mcp/commands.rs
+++ b/src/bin/caller/mcp/commands.rs
@@ -76,6 +76,9 @@ pub(crate) async fn start_task_with_state(
                     LogLevel::Info,
                     format!("Follow-up submitted via {}: {}", source, task),
                 );
+                // Dispatch changes process reality (the loop resumes work):
+                // drop any pre-dispatch controller-loop sample.
+                s.note_live_session_lifecycle_change();
                 drop(s);
                 tx.send(FollowUpMessage::text(task_clone))
                     .await
@@ -110,6 +113,11 @@ pub(crate) async fn start_task_with_state(
         LogLevel::Info,
         format!("Task started via {}: {}", source, task),
     );
+    // Task dispatch changes process reality (a session/agent is about to
+    // spawn): a cached no-process sample from before the dispatch must not
+    // answer the first post-start status poll (which could otherwise
+    // commit the session to Done via the stale-controller check).
+    s.note_live_session_lifecycle_change();
 
     let bus = bus.clone();
     drop(s);
@@ -920,6 +928,13 @@ pub(crate) async fn handle_control_command_mcp(
             let persistent = persistent.unwrap_or(true);
             match request_loop_halt_marker(&loop_dir, persistent) {
                 Ok(()) => {
+                    // Bump the cache generation before re-collecting: a
+                    // concurrent pre-warm holding a pre-mutation sample
+                    // must not store over the fresh one below.
+                    state
+                        .read()
+                        .await
+                        .invalidate_controller_loop_raw_status_cache();
                     let data = collect_controller_loop_status_with_state(&loop_dir, state).await;
                     emit_control_result(
                         control_tx,
@@ -943,6 +958,10 @@ pub(crate) async fn handle_control_command_mcp(
             let loop_dir = controller_loop_dir();
             match clear_loop_halt_markers(&loop_dir) {
                 Ok(()) => {
+                    state
+                        .read()
+                        .await
+                        .invalidate_controller_loop_raw_status_cache();
                     let data = collect_controller_loop_status_with_state(&loop_dir, state).await;
                     emit_control_result(
                         control_tx,
@@ -962,6 +981,10 @@ pub(crate) async fn handle_control_command_mcp(
             let loop_dir = controller_loop_dir();
             match request_loop_intervention_marker(&loop_dir, &mode) {
                 Ok(intervention) => {
+                    state
+                        .read()
+                        .await
+                        .invalidate_controller_loop_raw_status_cache();
                     let mut data =
                         collect_controller_loop_status_with_state(&loop_dir, state).await;
                     add_controller_loop_intervention_report(&mut data, &intervention);
--- a/src/bin/caller/mcp/commands.rs
+++ b/src/bin/caller/mcp/commands.rs
@@ -1123,6 +1123,9 @@ pub(crate) async fn handle_control_command_mcp(
             if let Some(ref tx) = s.follow_up_tx {
                 let tx = tx.clone();
                 s.set_phase(Phase::Thinking);
+                // Follow-up dispatch resumes work: drop any pre-dispatch
+                // controller-loop sample.
+                s.note_live_session_lifecycle_change();
                 s.push_log(LogLevel::Info, format!("Follow-up via socket: {}", text));
                 drop(s);
                 if tx.send(FollowUpMessage::text(text)).await.is_err() {
--- a/src/bin/caller/mcp/controller_loop.rs
+++ b/src/bin/caller/mcp/controller_loop.rs
@@ -1629,16 +1629,13 @@ pub(crate) fn controller_loop_state_is_idle(status: &str) -> bool {
 }
 
 pub(crate) fn codex_app_server_process_tree_active(pid: u32) -> bool {
-    // The root-alive probe is a cheap syscall, while the descendants
-    // snapshot spawns `ps` and parses the whole system process table —
-    // and `_with_root` never reads the descendants when the root is
-    // alive (the common case on this per-wrapper status hot path). Only
-    // pay for the snapshot once the root is known dead.
-    if crate::platform::process_alive(pid) {
-        return true;
-    }
-    codex_app_server_process_tree_active_from_descendants(
-        crate::platform::process_descendants(pid),
+    codex_app_server_process_tree_active_with_root(
+        pid,
+        // Lazy: `_with_root` never advances the descendants iterator when
+        // the root is alive (the common case on this per-wrapper status
+        // hot path), and materializing the descendants spawns `ps` over
+        // the whole system process table — defer the spawn to first use.
+        std::iter::once_with(move || crate::platform::process_descendants(pid)).flatten(),
         crate::platform::process_alive,
         crate::platform::process_cmdline,
     )
--- a/src/bin/caller/mcp/controller_loop.rs
+++ b/src/bin/caller/mcp/controller_loop.rs
@@ -528,11 +528,12 @@ pub(crate) fn collect_controller_loop_status_for_mcp_state(
     loop_dir: &std::path::Path,
     state: &McpAppState,
 ) -> serde_json::Value {
+    let (_, generation) = state.probe_controller_loop_raw_status(loop_dir);
     let raw = collect_controller_loop_raw_status(loop_dir);
     if state.controller_loop_status_override.is_none()
         && mcp_state_controller_loop_dir(state) == loop_dir
     {
-        state.store_controller_loop_raw_status(raw.clone());
+        state.store_controller_loop_raw_status_at(generation, raw.clone());
     }
     finish_controller_loop_status(raw, Some((state, current_unix_timestamp_secs())))
 }
@@ -639,14 +640,15 @@ pub(crate) fn mcp_state_controller_loop_status(s: &McpAppState) -> serde_json::V
             // to three times per call (promote + active/stale checks), and
             // supervised gates poll it continuously. Enrichment always runs
             // against the live state, so folded session phases are never
-            // stale even on a cache hit.
-            let raw = s
-                .cached_controller_loop_raw_status(&loop_dir)
-                .unwrap_or_else(|| {
-                    let raw = collect_controller_loop_raw_status(&loop_dir);
-                    s.store_controller_loop_raw_status(raw.clone());
-                    raw
-                });
+            // stale even on a cache hit. Stores carry the probe generation:
+            // if an invalidation (lifecycle/marker mutation) lands between
+            // probe and store, the pre-mutation sample is discarded.
+            let (hit, generation) = s.probe_controller_loop_raw_status(&loop_dir);
+            let raw = hit.unwrap_or_else(|| {
+                let raw = collect_controller_loop_raw_status(&loop_dir);
+                s.store_controller_loop_raw_status_at(generation, raw.clone());
+                raw
+            });
             finish_controller_loop_status(raw, Some((s, current_unix_timestamp_secs())))
         })
 }
--- a/src/bin/caller/mcp/controller_loop.rs
+++ b/src/bin/caller/mcp/controller_loop.rs
@@ -574,21 +574,47 @@ pub(crate) fn controller_loop_status_has_live_owner_or_process(status: &serde_js
             > 0
 }
 
-pub(crate) fn mcp_state_session_source_for_id(s: &McpAppState, session_id: &str) -> Option<String> {
-    s.session_source_for_id(session_id)
-        .map(str::to_string)
-        .or_else(|| {
-            (s.session_id == session_id)
-                .then(|| s.active_session_source.clone())
-                .flatten()
-        })
-        .or_else(|| {
-            match resolve_persisted_start_target(&mcp_state_session_logs_home(s), session_id) {
-                PersistedStartTarget::External(target) => Some(target.source),
-                PersistedStartTarget::ExternalMissingResume { source } => source,
-                PersistedStartTarget::NotFound | PersistedStartTarget::NonExternal => None,
-            }
-        })
+pub(crate) fn mcp_state_session_source_for_id(
+    s: &mut McpAppState,
+    session_id: &str,
+) -> Option<String> {
+    if let Some(source) = s.session_source_for_id(session_id).map(str::to_string) {
+        return Some(source);
+    }
+    if s.session_id == session_id {
+        if let Some(source) = s.active_session_source.clone() {
+            return Some(source);
+        }
+    }
+    // The persisted fallback reads and identity-scans the whole session log;
+    // a status poller for a session whose source never lands in
+    // session_sources would otherwise pay that scan on every call. Memoize
+    // both resolvable outcomes — resolved sources into session_sources,
+    // known-non-external ids into a dedicated set. NotFound stays uncached:
+    // the session's log may materialize a moment later.
+    if s.session_known_non_external.contains(session_id) {
+        return None;
+    }
+    match resolve_persisted_start_target(&mcp_state_session_logs_home(s), session_id) {
+        PersistedStartTarget::External(target) => {
+            s.session_sources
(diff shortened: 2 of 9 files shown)


lovon-spec:
Review round applied (both reviews reconciled). Per-item disposition:

| Item | Disposition | What changed |
|---|---|---|
| 1a — over-cap prune wipes ALL hydration cursors | **FIXED** | Prune drops only the ended component's cursors (log-dir basename match with the store's prefix semantics). Plus a hydration backstop: no folded status for the requested session ⇒ any cursor is dropped and the replay is full — a leftover cursor can never skip the prefix and answer from absent state. Test: `over_cap_prune_drops_only_the_ended_components_cursors`. |
| 1b — hydrating a pruned ended session re-prunes mid-rebuild, records an EOF cursor against absent state, later answers with the unrelated current session | **FIXED** | `note_session_ended` never prunes during hydration replays — query-driven rebuilds stay resident. Regression test reproduces the exact scenario (over-cap prune → rehydrate → assert Done on first AND second query): `pruned_ended_session_rehydrates_to_done_and_stays_consistent`. |
| 1c — hydration freshness watermark | **FIXED** | Live status writes mark the session; a FULL replay never overwrites a marked session's status (log flush skew), while appended-tail replays still apply (new information by construction). Marks pruned with the component. Test: `full_replay_does_not_regress_live_observed_phase`. |
| 2a — invalidation completeness | **FIXED** | Raw-sample cache now invalidates at task/follow-up dispatch, observed SessionStarted/SessionEnded/TaskComplete/DoneSignal/Interrupted (live folds only — hydration replays are no-ops via the replay flag), and the control-command marker paths (halt/clear/intervene) alongside the marker tools. |
| 2b — pre-warm race reinstating a pre-mutation sample | **FIXED** | Cache carries a generation; invalidate bumps it, stores refuse when it moved. The unlocked pre-warm, the inline miss path, and the fresh command collections all store at their probe generation. Test: `raw_status_cache_store_respects_invalidation_generation`. |
| 3 — cursor identity vs same/longer-length replacement | **FIXED** | `SessionJsonlCursor` freezes a prefix-identity window (first `min(consumed, 4096)` bytes, first-8-bytes-of-SHA-256 — the message-search `SourceCursor` discipline) at first establishment; resume verifies it, mismatch heals with a full replay. Covers both get_logs and hydration cursors. Test: `same_length_replacement_is_detected_by_the_prefix_identity_window` (same total length, same boundary newline — only the hash catches it). |
| 4 — NonExternal negative-memo poisoning on read errors | **FIXED** | A `session.jsonl` read failure now resolves as NotFound-class (never negative-cached); only a successfully read log with no external identity memoizes as non-external. |
| 5 — subscription set validation + cap | **FIXED** | `subscribe` validates against the served vocabulary (`RESOURCE_URIS`, parity-pinned to `resource_definitions()` by `resource_uris_pin_resource_definitions`) and rejects unknown URIs — the set is thereby capped at the fixed resource count. |
| 6 — close-time debounce flush reachability (adjudication) | **VERIFIED — Codex right** | Ownership chain: `run_mcp_server` moves `bus.clone()` into `spawn_event_listener`; the task holds that `EventBus` (a `broadcast::Sender<AppEvent>`) for its whole life to serve deferred control commands; a broadcast receiver reports `Closed` only when every sender is dropped — our own clone keeps the channel open, so the Closed arm is unreachable in production. Fix: the ownership argument is pinned in a comment (arm kept as defense in depth), and the reachable in-process teardown seam — the Quit control command — now flushes pending notifications immediately instead of letting the process exit inside the debounce window. |

Also in this round: `test_state` now defaults `controller_loop_dir_override` to a never-existing path — the new invalidation seams made non-hermetic tests re-collect controller-loop reality from the real machine home on every lifecycle fold (one test measured 2.3s → 21.9s on a live box); tests that exercise loop-dir behavior keep setting their own overrides.

lovon-spec:
Round 2 applied on `d8a7c640` (delta from `f8a74659`). Per-item disposition:

| Item | Disposition | What changed |
|---|---|---|
| 1 (blocker) — boolean watermark suppresses the wrong direction | **FIXED** | `session_status_live_phases` records the live-observed phase; full replays apply against the lattice: replayed **terminal** facts (Done/Interrupted) always apply — the lossless log outranks a stale live mark whose `SessionEnded` the lossy broadcast dropped — while replayed non-terminal phases never overwrite a live-marked session (transient skew self-heals via the live lane or the next terminal fact; documented at the field and at the gate). Mirror-image test: `full_replay_applies_newer_terminal_phase_over_lossy_live_mark` (live=Thinking marked, log=Done ⇒ replay applies, and stays applied after the EOF-cursor hydration). The `hydration_replay` flag is now RAII (`HydrationReplayGuard`, reset-on-drop) — a panic mid-fold cannot leave the state stuck in hydrating mode. |
| 2 — missing invalidations | **FIXED** | `ControlMsg::FollowUp` dispatch, both `LoopError` fold paths, and the same-class `SafetyCapReached` (listener) now bump the raw-sample cache generation. |
| 3 (blocker) — read failure silently spawns a fresh session | **FIXED** | New `PersistedStartTarget::Unreadable` classification: the source memo treats it like NotFound (never negative-cached, may retry), but the explicit resume arm surfaces an error instead of falling through to the generic start path (which discards the requested id). Test: `unreadable_persisted_log_errors_instead_of_spawning_a_fresh_session` — resolve returns Unreadable, `start_task` returns the error, and the bus receives **nothing** (no StartTask/ResumeSession). |
| 4 — unbounded flush; Quit-close question | **FIXED / VERIFIED pre-existing** | Flush bounded by a 500 ms budget (`RESOURCE_NOTIFY_FLUSH_BUDGET`): an open-but-not-draining peer drops the remainder instead of parking the listener. Quit verification: `should_quit` has **no** service-closing consumer in the MCP shape — `run_mcp_server` waits solely on transport completion (`running.waiting()`), and that predates this PR (TUI-era flag; the TUI is gone). My change only added the flush-on-quit seam; noting "Quit does not close the stdio service" as a follow-up below rather than fixing it in a perf PR. |
| 5 — SEP-2164 data.uri | **FIXED** | The unknown-URI `invalid_params` error now carries `data: {"uri": …}`. |
| 6 — cursor doc overstates | **FIXED** | Claim scoped to exactly the frozen window: a same-or-longer-length rewrite diverging only beyond it (and preserving the boundary byte) is not detected — acceptable under the append-only writer invariant. |
| 7 — hermeticity threading | **FIXED (threaded)** | The wrapper-index home is now a parameter through `collect_controller_loop_raw_status` → `active_external_wrappers_from_index_for_processes` → `controller_loop_wrapper_index_homes`; state-scoped callers pass the test-overridable `mcp_state_session_logs_home`, stateless transport edges (`collect_controller_loop_status_inner`) resolve the real home at their boundary per the convention. `test_state` defaults both roots (loop dir + logs home) to never-existing paths. Honest residual, documented at the fixture: one `ps` probe of this process's descendants remains (platform probe, not state) — deterministically empty for test processes, and even a match can no longer read the real home's index. |

**Follow-up (out of scope, pre-existing):** `ControlMsg::Quit` sets `should_quit` but nothing in the stdio MCP shape closes the service on it — `run_mcp_server` returns only when the transport ends. Worth its own small change if Quit is meant to terminate the process.

Battery (`CARGO_PROFILE_DEV_DEBUG=0`): `cargo fmt --check` clean; `cargo clippy` (workspace) 0 warnings; `cargo nextest run --bins` 3492 passed / 0 failed; `cargo test --test e2e` 21 passed / 0 failed.

lovon-spec:
Round 3 applied on `76f0d835` (Option A, as adjudicated): the phase-kind watermark is deleted — `session_status_live_phases`, both lattice gates, and `phase_is_session_terminal` are gone. `hydration_replay` (RAII) stays for prune suppression + cache-invalidation gating only (replay-vs-reality concerns, not ordering). Replayed rows now apply ungated and the fold lands on the log's final state — permanently correct in all four reviewed directions, with the transient sink-lag window documented at the hydrate fold site. The emission-stamped ordering token (Option B) and the pre-existing Quit-does-not-close-the-service behavior are filed in the PR body's new Follow-ups section.

Tests pin the new semantics: `full_replay_folds_through_stale_terminal_rows_to_the_logs_final_state` (the resumed-after-Done case the lattice broke), `delayed_tail_rows_apply_and_self_heal_when_the_log_catches_up` (documented transient + heal; the heal rides a session-id-carrying row since round_complete rows persist no session id), and `full_replay_applies_a_missed_terminal_fact_from_the_lossless_log` (renamed round-2 case, unchanged behavior). ITEMs 2-7 untouched.

Battery (`CARGO_PROFILE_DEV_DEBUG=0`): fmt --check clean; clippy (workspace) 0 warnings; `cargo nextest run --bins` 3493 passed / 0 failed; e2e 21 passed / 0 failed.

lovon-spec:
Round 4 applied on `4de3a98b` (replay-attribution blocker):

**ITEM 1 — FIXED in the MCP reducer, session_log/** untouched.** The hydration-replay guard now carries the TARGET session id; `unattributed_event_session_id()` resolves a missing session id to the hydration target while a replay is in scope — never to `self.session_id` — and live folding keeps the historical active-session fallback unchanged. All three reducer fallback sites route through the helper (`note_session_phase`, `note_session_round`, and `rewind_session_key` — the same unattributed-event question). The unread `HydrationReplayKind` (Full/Tail) is gone: nothing branches on it since the round-3 lattice removal; the guard's payload is now the target id. Regression test per the required shape — `hydration_attributes_unattributed_rows_to_the_target_session` uses the REAL `SessionLog::round_complete` writer with the naked row as the log's FINAL line (no attributed bracket after it) and an unrelated active session: requested session reaches WaitingFollowUp (round 2), active session untouched, both stable once the cursor sits at EOF. The round-3 agent_started-bracket heal test stays green alongside.

**ITEM 2 — filed.** PR body follow-ups now include stamping `session_id` on `round_complete` rows at the writer (source fidelity; lets replay reconstruct fully-attributed events) — deferred because it touches `session_log/**`, inside PR #344's in-flight diff.

Round-3 fold semantics and ITEMs 2-7 untouched. Battery (`CARGO_PROFILE_DEV_DEBUG=0`): fmt --check clean; clippy (workspace) 0 warnings; `cargo nextest run --bins` 3494 passed / 0 failed; e2e 21 passed / 0 failed.

lovon-spec:
Round 5 applied on `ecfb822a`: the hydration bracket now snapshots/restores `s.round` alongside `s.turn` — a replayed `round_complete` can no longer leak the old log's round into the daemon-global scalar that unscoped `get_status` reports. The round-4 regression now initializes a distinct live round (9) and asserts BOTH the global `s.round` and the live session's stored round after the first and the EOF rehydration (the prior version asserted only the live phase, which is why the defect slipped it).

Bracket completeness audit (now pinned as a comment at the bracket): enumerated all 46 AppEvents the replay converter can produce, intersected with the applier arms that write daemon-global scalars — {SessionIdentity, SessionCapabilities, ContextSnapshot, SessionStarted, TurnStarted, AgentStarted/AgentOutput, DoneSignal/TaskComplete, ApprovalRequired, RoundComplete, SessionEnded}; every scalar they write (turn, round, budget_pct, phase + phase_entered_at, task_description, session_id, active_session_source, codex_managed_context, provider/model names, the five token counters, context/hard windows) is in the bracket. Globals the applier CAN write but no replayable row reaches — presence_* (UsageSnapshot isn't producible; Live/PresenceUsageUpdate hit the applier's `_ => false`), external_agent, configured_codex_managed_context, log_dir — are safe by unreachability and flagged in the comment for re-audit if the converter grows a producer; approval/human-question/log-ring state is never written by the applier at all (verified: no push_log in the applier).

Battery (`CARGO_PROFILE_DEV_DEBUG=0`): fmt --check clean; clippy (workspace) 0 warnings; `cargo nextest run --bins` 3494 passed / 0 failed; e2e 21 passed / 0 failed.

PR #512, opened 2026-07-19
steers: reconcile fallback acks across id aliases; re-enable CC mid-turn injection

lovon-spec:
## What

Two fixes from the live e2e report "native steer was not acknowledged; queued as follow-up" (steer to a Claude Code session addressed by its backend id):

1. **Fallback ack reconciliation** (`session_supervisor/routing.rs`): drains normalize backend-native steer targets and ack under their primary wrapper id (`normalize_native_session_target` → `resolve_external_steer_target_session`), but the 2s text-steer fallback only matched acks against the requested id form. Every alias-addressed steer was therefore parked as a duplicate follow-up and its honest queue status overwritten by a false "not acknowledged". The fallback now accepts an ack under either the requested alias or the resolved managed id (new `steer_ack_targets_session_or_resolved`), with a regression test.

2. **Claude Code mid-turn injection re-enabled** (`external_agent/claude_code.rs`): `steer_turn` writes the stream-json user message onto stdin instead of refusing with "mid-turn steering not supported". Verified live on CC 2.1.215 (twice, haiku): a message written during a running turn is absorbed into that same turn at the CLI's next checkpoint; the CLI never echoes it on stdout, so delivery is inferred via the existing pending-runtime-steer checkpoint tracking (turn completion at the latest). The 2.1.207-era discard behavior is documented as history; the idle path ("no active turn" → immediate follow-up) is unchanged.

Docs/comments refreshed to the verified behavior: capability table + steer section in `docs/src/external-agent-orchestration.md`, goal-notice rationale (prelude path deliberately kept — notices must never vanish on a discard-era CLI), and the live-e2e SKILL troubleshooting notes.

## Verification

- `cargo fmt --check`, `cargo clippy --workspace -- -D warnings`
- `cargo nextest run -p intendant --bins` (4503 passed)
- lib crates + `cargo test -p intendant --test e2e` (24 passed)
- Live raw-CLI probes on 2.1.215 (haiku): mid-turn stdin message absorbed into the running turn's final reply; no stdout echo of the injected message
- Live `tests/skills/claude-code-e2e` driver run: see PR comment

### DIFF
--- a/docs/src/external-agent-orchestration.md
+++ b/docs/src/external-agent-orchestration.md
@@ -110,7 +110,7 @@ adapter from `[agent.<backend>]` config, then `run_external_agent_mode()`
 | MCP injection | Per-process `-c mcp_servers.intendant.{type,url,bearer_token_env_var}` overrides plus scoped env; no workspace config file | Inline `--mcp-config '{…}'` JSON with an environment-expanded Authorization header |
 | Multi-thread | Yes — many threads per process | No |
 | Native thread id | Yes | Yes — announced via `AgentEvent::NativeSessionId` on the first turn (placeholder `claude-code-session` until then; `--resume` keeps the id stable so resumed threads are canonical immediately) |
-| Mid-turn steer | Yes (`turn/steer`) | No — 2.1.2xx discards stdin user messages mid-turn (2.1.200's absorb was a CLI bug, since removed); `steer_turn` reports queue semantics and the steer delivers at the turn boundary (or immediately as its own turn when idle) |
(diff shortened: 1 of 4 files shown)


lovon-spec:
Live `tests/skills/claude-code-e2e` driver run (haiku, debug build, CC CLI 2.1.215): **20/21 checks passed**, including the acceptance check for this PR — `steer-absorbed-in-turn` ("steered.txt created within the steered turn", i.e. the mid-turn stdin write was absorbed by the running turn through the full control-socket → routing → drain → `steer_turn` path). Approvals, deny, native-id, usage, capabilities, interrupt (+process survival), goals, compact, and fork spawn/identity all passed.

The one failure is environmental: `scenario-completed` timed out waiting for fork context recall because the account's Claude **five_hour rate limit tripped mid-run** at +220s ("status rejected (five_hour window) — resumes 7:20 PM"); the fork child parked on the limit (the rate-limit park + auto-resume machinery engaged correctly) and the driver's recall window expired. Unrelated to this diff — no fork/resume code is touched.

PR #521, opened 2026-07-20
steers: confirm delivery on tool-call checkpoints; waiting rows show elapsed

lovon-spec:
Follow-up to #512, from live e2e: a mid-turn steer to a Claude Code session sat in "waiting for the next runtime checkpoint" for 5 minutes while the model chained tool calls without emitting text — the steer was already absorbed (it produced an agenda item mid-turn), but delivery confirmation only fires on model text, so the row read as stuck.

1. **Delivery confirms on tool-call checkpoints** (`external_events.rs`): the `ToolStarted` arm now calls `mark_pending_runtime_steers_delivered_at_model_checkpoint` — a model-issued tool call is the same fresh-API-response evidence as streamed text. The helper's existing session filter keeps a Task sub-agent's tool calls from confirming parent steers.
2. **Honest accepted wording**: `SteerAccepted` reason becomes "<agent> injected the message into the running turn — awaiting the model's next activity".
3. **Elapsed clock on waiting rows** (`54-session-lifecycle.js` + CSS): pending/accepted/queued steer rows show `· 1m 23s`, driven by a single shared 1s ticker that starts with the first waiting row and self-stops with the last; cleared on terminal statuses. `static/app.html` regenerated via the assembler.

## Verification
- `cargo fmt --check`, `cargo clippy --workspace -- -D warnings`
- `cargo nextest run -p intendant --bins` (4516 passed), lib crates, `--test e2e` (24 passed)
- `node --check` on the fragment; live `tests/skills/claude-code-e2e` driver run: see PR comment

### DIFF
--- a/src/bin/caller/event.rs
+++ b/src/bin/caller/event.rs
@@ -265,8 +265,9 @@ pub enum AppEvent {
     SteerAccepted {
         session_id: Option<String>,
         id: String,
-        /// Short human-readable detail for UI display, e.g. "Codex accepted
-        /// the steer; waiting for the next runtime checkpoint".
+        /// Short human-readable detail for UI display, e.g. "Codex injected
+        /// the message into the running turn — awaiting the model's next
+        /// activity".
         reason: String,
     },
     /// Mid-turn steering could not be delivered natively by the current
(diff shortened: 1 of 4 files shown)


lovon-spec:
Live `tests/skills/claude-code-e2e` driver run (haiku, debug build, CC CLI 2.1.215): **all checks passed** — full board including `steer-absorbed-in-turn` with the new tool-call confirmation path in place, approvals, interrupt, goals, compact, fork (+context recall), Task sub-agents, and resume-binds-fork.

PR #586, opened 2026-07-24
Add Codex Cloud worker leases

lovon-spec:
## What

- add `intendant codex-cloud` commands for provider health, task submission, lease refresh/status, task diffs, attachment state, and bootstrap generation
- model Codex Cloud tasks as ephemeral worker leases with independent provider and live-attachment lifecycles
- expose full-profile MCP tools to list workers and submit Cloud tasks through the daemon host's authenticated Codex CLI
- add setup, cached-environment maintenance, and task-time foreground runner scripts with fresh per-task identity state
- document the lifecycle, credential boundaries, and current broker boundary

## Why

Codex Cloud containers are provider-owned and may outlive a terminal task briefly or disappear without an orderly peer disconnect. Treating them as permanent `[[peer]]` entries would conflate job state with reachability and risks reusing identity from a cached environment.

Provider subprocesses also run in disposable working directories because the upstream CLI currently writes an account-bearing `error.log` in its current directory.

## Safety / impact

- no AWS, relay, DNS, security-group, public-port, or production app changes
- no relay credentials, private keys, static peer identity, or fixed routes are embedded
- setup-time secrets are not persisted into the agent phase
- the live enrollment broker and multi-tenant route allocator remain future work; the generated runner accepts an explicit foreground supervisor

## Checks

- `cargo build --bin intendant`
- `cargo test --bin intendant codex_cloud::tests -- --nocapture` (10 passed)
- `cargo test --bin intendant list_tools_ -- --nocapture` (3 passed)
- `bash -n scripts/codex-cloud/*.sh`
- `mdbook build docs`
- `git diff --check`
- authenticated read-only `codex-cloud doctor` and `list` smoke tests
- generated bootstrap and task-local runtime smoke test
- full current-base binary suite: 4,857 passed, 10 ignored, 0 failed


lovon-spec:
I reviewed this end to end. My recommendation is to **keep this PR in draft and address the findings below before merge**. The overall architecture is promising, but there are four correctness issues I think should block readiness.

### Findings

1. **[P1] The fingerprint probe cannot detect cold replacement as advertised**

   The lease stores only one fingerprint. Once populated, refresh deliberately stops fetching the diff. `probe` also creates a new task rather than another turn of the same task, so it cannot compare worker identity across turns. The documentation's claim that “mismatches across turns prove a cold replacement” is therefore not implemented by the current instrument.

   Relevant code: [fingerprint collection](https://github.com/intendant-dev/Intendant/blob/67b06323ef0c93f68c85600aa68817ae51ad4066/src/bin/caller/codex_cloud.rs#L650-L662), [probe submission](https://github.com/intendant-dev/Intendant/blob/67b06323ef0c93f68c85600aa68817ae51ad4066/src/bin/caller/codex_cloud.rs#L803-L855).

   I would store fingerprint history keyed by turn, write each fingerprint to a unique path, and collect after every observed terminal edge.

2. **[P2] Follow-up authentication can use a stale token**

   `auth.json` is loaded before `ensure_task_idle()`. That idle check runs the Codex CLI, which may refresh and rewrite the token, but the subsequent private API requests still use the old in-memory credentials. The first follow-up after expiry can therefore fail with 401 while a retry succeeds.

   Relevant code: [auth and follow-up ordering](https://github.com/intendant-dev/Intendant/blob/67b06323ef0c93f68c85600aa68817ae51ad4066/src/bin/caller/codex_cloud.rs#L1264-L1320).

   Load credentials after the CLI check. `doctor` should also detect unsupported keyring-backed Codex logins: upstream Codex supports file, keyring, auto, and ephemeral credential stores, whereas this path reads only `auth.json`.

3. **[P2] A successful follow-up to an old, untracked task is not recorded**

   Tasks outside the newest 20 are explicitly accepted through `codex cloud status`, but after the backend accepts the turn, the running edge is updated only if the lease already exists. Consequently, that invocation can report success while leaving no lease, warmth, or turn-tracking state.

   Relevant code: [outside-window acceptance](https://github.com/intendant-dev/Intendant/blob/67b06323ef0c93f68c85600aa68817ae51ad4066/src/bin/caller/codex_cloud.rs#L1207-L1249), [conditional update](https://github.com/intendant-dev/Intendant/blob/67b06323ef0c93f68c85600aa68817ae51ad4066/src/bin/caller/codex_cloud.rs#L1354-L1367).

   Either upsert a minimal lease after acceptance or refuse untracked tasks.

4. **[P2] The downloaded-binary cleanup trap references an expired local variable**

   `downloaded` is function-local, while the single-quoted `EXIT` trap expands it after the function returns. Under `set -u`, Bash reports `downloaded: unbound variable`, and the temporary download remains.

   Relevant code: [setup.sh](https://github.com/intendant-dev/Intendant/blob/67b06323ef0c93f68c85600aa68817ae51ad4066/scripts/codex-cloud/setup.sh#L17-L21).

### Design input

The strongest parts are the provider/attachment state separation, disposable Codex working directories, task-local Intendant identities, fail-closed handling of private API drift, and pulling results into a fresh worktree. Those are good foundations.

The runtime documentation should also incorporate the later controlled cold-resume result: after roughly 34 quiet minutes, a same-task follow-up received a different worker and lost its ignored Cargo target, `/root`, `/tmp`, and background-process state. That means an external build cache is a requirement for dependable reuse, not merely an optimization.

Validation I ran against commit `67b06323`:

- `cargo test --bin intendant codex_cloud::tests -- --nocapture`: **32 passed**
- `git diff --check`: passed
- all three shell scripts pass `bash -n`
- no GitHub checks are currently attached

The PR also currently conflicts with `main` in generated `static/app.html`. Per the repository's no-history-rewrite rule, I recommend merging current `main` into the branch and resolving/regenerating there rather than rebasing.

PR #740, opened 2026-08-02
closable lens legibility: on-canvas direction, auto-disengage, phase-guarded × title

lovon-spec:
Commissioned from agenda item 01KYXD2VNF6XT4SCA9282H75WS (occurrence 6b506af11914865084c5aba883dc505d): the Activity closable machinery is correct, but its presentation inverted the owner's read — a returning viewer saw dimmed running/parked cards and read dim as *closable* (the opposite is true), and the agenda-settled × tooltip read like an all-clear on a running card. Display only; zero semantics change — `sessionWindowClosableClaim` and the Arrange walk are untouched.

- **On-canvas direction while engaged**: the chip's own text flips to `N safe to close · rest dimmed`, and a grid legend (keyed on the same `data-ui2-closable-lens` attribute as the dim, so it can never outlive the lens) restates which side is which, with a **Show all** disengage button. Hover is never the only door.
- **Stickiness reconsidered → auto-disengage**: navigating away from the Timeline grid (main tab, Activity sub-tab, or Focus-layout flip) disengages the lens — the forgotten-lens re-entry that produced the specimen can no longer happen. Doctrine comment extended: the lens is a look, not a mode.
- **Phase-guarded settled × title**: agenda-linked settled + non-quiet window now leads with the live pill label — `Running Agent — stopping interrupts it; no agenda-owed work remains (the linked occurrence is settled)` — preserving two-axis honesty. Quiet mirrors the lens claim's rule (hard done, or a present phase normalizing to idle); an absent phase keeps the agenda-axis copy. The title now re-derives on **every** phase application: the phase-only fast path skips the wide render, which previously let the old titles go stale across phase flips.
- **QA vectors**: `window.qa.closableLens` gains `closeTitle()`, `chipCopy()`, and `vectors()` — a self-checking pinned matrix over the × title claims and the chip/legend copy — plus `chipWord`/`chipTitle`/`legendShown` in `state()`. Rust-side pin test `closable_lens_states_its_direction` added beside the existing positive-only pin; `tests/skills/closable-lens-qa/SKILL.md` gains the copy probe.

Gate question 'Gate: closable lens legibility' will be parked at queue entry.

### DIFF
--- a/src/bin/caller/web_gateway/static_assets.rs
+++ b/src/bin/caller/web_gateway/static_assets.rs
@@ -2365,4 +2365,28 @@ mod tests {
         );
         assert!(APP_HTML.contains("win.el.classList.toggle('session-window-closable', closable)"));
     }
+
+    /// The engaged lens states its direction ON-canvas (live specimen
+    /// 2026-07-31: a returning viewer read the dim as "closable" — the
+    /// exact inversion): the chip's own text flips to name the bright
+    /// side, the grid legend restates it keyed on the same html
+    /// attribute as the dim (it can never outlive the lens), navigating
+    /// away from the Timeline grid disengages the look, and the
+    /// agenda-settled × title is phase-guarded — on a non-quiet window
+    /// it leads with the live pill label instead of reading as an
+    /// all-clear, re-derived on every phase application because the
+    /// phase-only fast path skips the wide render.
+    #[test]
+    fn closable_lens_states_its_direction() {
+        assert!(APP_HTML.contains("safe to close · rest dimmed"));
+        assert!(APP_HTML.contains("id=\"ui2-closable-lens-legend\""));
+        assert!(APP_HTML.contains(
+            "html.ui-v2[data-ui2-closable-lens=\"on\"] .ui2-closable-lens-legend { display: flex; }"
+        ));
+        assert!(APP_HTML.contains(
+            "— stopping interrupts it; no agenda-owed work remains (the linked occurrence is settled)"
+        ));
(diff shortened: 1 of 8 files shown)


lovon-spec:
Validation (local, anchored in the worktree):

**Keyless battery** — `cargo fmt --check` 0, `cargo clippy --workspace -- -D warnings` 0, `cargo test -p intendant --bins` 0 (5488 passed incl. `closable_lens_is_positive_only` + new `closable_lens_states_its_direction`), lib-crate suite 0.

**Browser QA leg** (validate-dashboard.cjs, real Chrome, scratch `INTENDANT_HOME`, worktree debug binary; grid layout applied first — fresh profiles boot in Focus where the lens surface is CSS-gated):
```
probe boot = {"layout":"grid"}
probe copy = {"rows":17,"pass":true,"failed":[]}
probe lens = {"offWord":"closable","offLegend":false,"onWord":"safe to close · rest dimmed","onLegend":true,"onAttr":"on","onTitleLeads":true,"offAgainLegend":false}
probe title = {"runLeads":true,"idleSettled":true,"flipTracks":true,...}
probe disengage = {"afterSubtab":false,"afterTab":false,"backOn":false,"afterFocus":false}
probe stage = {"on":true,"word":"safe to close · rest dimmed","legend":true,"count":1}
```
`title.flipTracks` proves the phase-only fast-path staleness fix live (settled title tracks a running→idle flip that skips the wide render); `disengage` drives real nav clicks. Engaged-state screenshot held locally as gate evidence.

PR #751, opened 2026-08-02
Update abstraction slice 1: live-holder doorway resolution + tokened doorways (+R-D1)

lovon-spec:
Implements **update-abstraction slice 1** per the sealed intake (§7 slice 1, ruled PASS 2026-08-01) and its ruling tail, from commission card 01KZ0189S8R1PABK9ZXX2CWZSN.

## What changes

- **`resolveLiveHolder(body)`** (ui2-handover.js): the §5.1 rule — prefer the lease-sidecar target **iff its daemons entry is `live && state !== 'draining' && state !== 'exited'`**; else fall back to the existing hand-off candidate rule (`handoverSuccessorCandidate`: live, non-draining, on-disk build preferred); else null. The drain banner's successor doorway now consumes this resolver, so the 6c entry→acquisition window (sidecar naming the most recent **drainer**) can never render a doorway into a drain — it renders the honest "No successor has acquired the lease yet." instead.
- **Tokened doorways**: `handoverDaemonLink` (successor *and* predecessor doorways) decorates the URL at click time through the same-home sibling token map (`withSiblingLoopbackToken`), so on the loopback posture the click lands authed instead of on the named 401. The helper is **lifted** from `43-access-panes.js` into `31-init-identity-fleet.js` beside the loopback-token machinery it extends (shared with the Access fleet links — never duplicated). Modifier/middle clicks keep native semantics on the bare href.
- **R-D1 — implemented** (not deferred): `HandoverRuntime::successor_port()` is now the server-side mirror of the same rule for `daemon_draining` refusal pointers — the sidecar target is presence-consulted (live, not draining/exited) before being named; a live non-draining sibling is the fallback; None ("successor pending") otherwise. Scope note: the mirror carries the *eligibility* rule; the on-disk-build **preference** stays client-side where `update.on_disk` is at hand — it is a doorway product refinement, not a refusal-pointer concern (any live non-draining daemon serves refused work equally).

## Acceptance mapping (gate criteria, intake §7 slice 1)

- **(a) fragment needle pins, same commit**: `doorway_resolves_live_holder_and_carries_the_sibling_token` pins the resolver, the filter expression (`entry.live && entry.state !== 'draining' && entry.state !== 'exited'`), the doorway consumption, the fallback composition, the click-time token decoration, the lift (helper present in 31, absent from 43, exactly once in the served bundle), and the removal of the boot-id-only `handoverSuccessorPort`. `successor_port_never_names_a_draining_or_dead_target` pins the server mirror across the full 6c chain (spare fallback → sidecar arm → draining window → healed sidecar → crashed successor).
- **(b) e2e three-daemon chain leg**: `chained_drain_payload_resolves_only_live_non_draining_holder` (spawn_co_daemon ×3, `INTENDANT_LEASE_POLL_MS=500`, the :6835/:6921 rig family) — A drains toward B; B drains with no successor (the 6c window, held open by parked holdout sessions on both drainers): A's served payload carries sidecar={B, draining} with B live+draining in the daemons array and the resolution rule refuses it; C boots, acquires, and the same payload lets the rule reach C while B still shows live+draining beside it. `/api/local-daemons/tokens` on the drainer names the successor's own admission token, and the token admits the exact doorway URL shape (`/?token=…` → 200). The drainers' refusal pointers ride the mirror end to end: "has not acquired" during the window (never :B), "(:C)" after. Holdouts park on a long provider delay rather than the 30s-bounded file barrier — two takeovers plus a third boot don't fit under 30s on a slow runner.
- **(c) full local battery**: bins + lib crates + e2e + workspace clippy -D warnings + fmt --check — results will be posted on this PR before it leaves draft.

## Bindings held

No new authority: the doorway stays client navigation; the token map route (`/api/local-daemons/tokens`, owner-posture, CredentialsManage class) is consumed exactly as the fleet links already consume it; no tunnel twins added (P3(b) intact). Era vocabulary untouched. F1 (presence watershed) untouched — `successor_port` only reads presence.

### DIFF
--- a/src/bin/caller/handover/mod.rs
+++ b/src/bin/caller/handover/mod.rs
@@ -1231,7 +1231,8 @@ mod tests {
             );
         }
         assert_eq!(
-            app.matches("async function withSiblingLoopbackToken").count(),
+            app.matches("async function withSiblingLoopbackToken")
+                .count(),
             1,
             "one shared sibling-token helper in the served bundle"
         );
@@ -1512,7 +1513,11 @@ mod tests {
         // still names A itself — the fallback names the live spare
         // rather than leaving callers with "pending".
         assert_eq!(a.request_drain(None), DrainRequest::Entered);
-        assert_eq!(a.successor_port(), Some(7002), "a live spare beats 'pending'");
+        assert_eq!(
+            a.successor_port(),
+            Some(7002),
+            "a live spare beats 'pending'"
+        );
 
         // B acquires: the sidecar arm names it.
         assert!(b.poll_acquire(0));
@@ -1534,7 +1539,11 @@ mod tests {
         // C crashes: the sidecar still names it, but a dead boot is not
         // a doorway either (B stays draining, so nothing is eligible).
         drop(c);
-        assert_eq!(a.successor_port(), None, "a crashed successor is not a doorway");
+        assert_eq!(
+            a.successor_port(),
+            None,
+            "a crashed successor is not a doorway"
+        );
     }
 
     /// The Q4 pin: a drainer that observes its successor gone — crashed
--- a/src/bin/caller/handover/mod.rs
+++ b/src/bin/caller/handover/mod.rs
@@ -724,12 +724,38 @@ impl HandoverRuntime {
         )
     }
 
-    /// The successor's port for `daemon_draining` refusal pointers, when
-    /// the sidecar already names one (None while the handoff is still in
-    /// flight — the refusal then says "successor pending").
+    /// The successor's port for `daemon_draining` refusal pointers — the
+    /// server-side mirror of the dashboard's `resolveLiveHolder` rule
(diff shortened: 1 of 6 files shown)


lovon-spec:
Battery on the merged tree (post `origin/main` @ 96e4e5d6 reconcile — successor-exec #743 landed under this PR; sole conflict was generated app.html, resolved by the assembler ritual):

- `cargo fmt --check` — clean
- `cargo test -p intendant --bins` — 5532 + 150 + 97 passed, 0 failed
- lib crates (`intendant-core`, `intendant-custody`, `intendant-display`, `intendant-platform`, `owner-plane-core`, `owner-plane-reducer`) — 14 suites, all ok
- `cargo test -p intendant --test e2e` — 44/44 (incl. the new `chained_drain_payload_resolves_only_live_non_draining_holder` leg and #743's successor-exec legs)
- `cargo clippy --workspace -- -D warnings` — clean

PR #781, opened 2026-08-03
dashboard: auto-follow + carry across the daemon swap (update-abstraction slice 2)

lovon-spec:
## Update-abstraction slice 2 — auto-follow + carry (browser surfaces)

Implements intake §7 slice 2 (sealed: ~/update-abstraction-intake.md, sha256 ed4215e8…) with the ruling's four refinements: **R-A1** (completion honesty compares the git_sha+built_at STAMP PAIR, never sha alone), **R-A2** (the sibling token map is re-fetched, not only pre-fetched — and the same re-fetch-on-miss heals the slice-1 doorway's per-page-load cache, the ruling's N1 note, in this same change), **R-A3** (an over-cap composer draft names its loss BEFORE navigation — consent, not disclosure), **R-A4** (the legacy `[server.auth] bearer_token` posture: the bearer **rides the carry blob** — it is already a client-held per-origin secret and the capture-strip pattern applies verbatim; chosen over the honest-copy degrade so LAN-bearer surfaces keep working across the hop).

Composes with slice 1's landed substrate (resolveLiveHolder + tokened doorway, PR #751) — nothing duplicated: the follow watch consumes the same resolver (pinned: ONE rule, two consumers) and the same token helper.

### What ships
- **Follow watch** (ui2-handover.js): arms on own-drain; 2.5s tick re-fetches the drainer's payload, re-resolves the live holder (never onto a draining daemon — 6c), fetches the target's admission token via the shared map helper, probes the target origin (no-cors HEAD, the BackendSupervisor readiness pattern), then `location.replace` with `?token=`, the route hash, and the carry. A drainer that stops answering after the target was resolved+probed is itself the follow signal.
- **Carry blob** (§2.3): `#…&carry=<base64 json>` — route, open session-window SET, the two sub-tab keys, pre-swap build stamps, optional bearer (R-A4), optional consented draft; ~8KB cap, draft drops first. Landing side (31-init-identity-fleet.js) capture-strips it before the router/session-windows readers evaluate and seeds their own persisted keys — the successor rebuilds windows from its own catalog.
- **Q1 posture**: hidden tabs move immediately; visible tabs after a ~5s grace toast ('Moving to the updated daemon…'); any composer draft demotes to a one-button 'Continue on the updated daemon' consent (pending consent also outranks the banner's stored collapse).
- **Completion honesty** (§2.4 + R-A1): the landing page compares its own presence-entry stamps against the carried predecessor stamps as the PAIR; 'All set — running <version>.' only on a real change, 'The restart didn't change builds.' otherwise, neutral copy when stamps are unreadable.
- **Exclusions** (§2.5): packaged-app webview stands down silently (the app follows itself via didSwapToPort); hosted-relay surfaces get honest banner copy — port substitution means nothing there; canonical-port rebind stays the parked durable fix for cold bookmarks and relay surfaces.
- **QA**: `qa.followState` (token presence only, never the token), `qa.followCarryApplied`, and the `__testFollowDecide` driver (the `__testNudgeDaemonBoot` precedent).

### Acceptance (intake §7 slice 2, verbatim) — all four met
- **(a)** e2e `follow_contract_carries_the_scripted_tab_to_the_successor`: the served contract walked as the scripted tab — mid-window resolution None + the pre-fetch floor that cannot name an unbooted successor; after C acquires, re-resolution reaches :C, the RE-fetched map names C's own admission token (R-A2 pinned end to end), the origin answers the probe, the tokened follow URL lands an authed 200, and both payloads carry the stamp-pair compare substrate.
- **(b)** four fragment pins in `handover/mod.rs`: `spa_arms_the_follow_watch_on_own_drain`, `follow_carry_capture_strips_and_seeds_the_landing` (incl. the sub-tab literal parity pin and served-bundle exactly-once counts), `follow_composer_guard_demotes_to_consent` (R-A3), `completion_honesty_compares_the_stamp_pair` (R-A1).
- **(c)** `tests/skills/handover-follow/SKILL.md` — headed scenario, **NOT in CI** (stated in the doc): probe leg (validate-dashboard + qa.handoverBanner + qa.followState + __testFollowDecide) and the live two-daemon eyes-on follow leg whose same-binary completion line must read 'The restart didn't change builds.' **Probe leg executed live at this head**: `followArm`/`followConsent`/`followGrace` all-true, PASS, exit 0 (pipefail).
- **(d)** full local battery, bare exits all 0: `cargo fmt --check`; bins 5616+150+97; lib crates (6 packages); e2e **47/47** incl. the new leg; `cargo clippy --workspace -- -D warnings`.

### Trust posture
No new routes, no tunnel twins, no new authority: the follow is pure client-side navigation plus the existing owner-posture token map, consumed exactly as the fleet links already consume it. `gateway_routes.rs` untouched. The carry is the same trust class as the existing `?token=` capture — everything in it is data a linking page could already put in a URL; junk seeds break only the tab's own display state.

### Notes for the gate
- The relay exclusion is detected via `hostedControlActive()` ∪ `DASHBOARD_CONNECT_MODE` (the latter a build-time constant today — future-proofing, zero cost).
- Slice-3 seat (s-4ab14652) coordinated on the bus: disjoint fragments, disjoint pin files, e2e legs anchored apart; app.html conflicts resolve in fragments + assembler.

### DIFF
--- a/src/bin/caller/handover/mod.rs
+++ b/src/bin/caller/handover/mod.rs
@@ -1245,6 +1245,172 @@ mod tests {
         );
     }
 
+    /// Update-abstraction slice 2 pin (intake §7, acceptance (b)): the
+    /// follow watch — armed exactly by the own-drain render branch,
+    /// resolving through the ONE live-holder rule (the doorway's — two
+    /// consumers, one rule, never a second), tokening through the
+    /// shared sibling map with R-A2's re-fetch-on-miss, probing the
+    /// target origin (no-cors HEAD), and moving the tab with
+    /// location.replace + the tokened URL + the carry fragment. The
+    /// packaged-app webview stands down; a relay surface gets honest
+    /// copy instead of a meaningless port substitution.
+    #[test]
+    fn spa_arms_the_follow_watch_on_own_drain() {
+        let fragment = include_str!("../../../../static/app/ui2-handover.js");
+        for needle in [
+            "function armFollowWatch",
+            "armFollowWatch(body);",
+            "function followWatchTick",
+            "function followSurfaceExclusion",
+            "siblingTokenForPort(followWatch.port)",
+            "mode: 'no-cors'",
+            "method: 'HEAD'",
+            "location.replace(url)",
+            "&carry=",
+            "Moving to the updated daemon…",
+            "if (document.hidden) {",
+            "FOLLOW_GRACE_MS",
+            "through a relay",
+            "if (followWatch.armed) disarmFollowWatch();",
+            "window.qa.followState",
+        ] {
+            assert!(
+                fragment.contains(needle),
+                "the follow watch lost its wiring: {needle}"
+            );
+        }
+        // ONE resolution rule, two consumers: the doorway and the watch
+        // tick both read the holder through resolveLiveHolder — a
+        // follow that resolved any other way could walk into a drain.
+        assert_eq!(
+            fragment
+                .matches("const holder = resolveLiveHolder(body);")
+                .count(),
+            2,
+            "doorway + follow tick both consume the one live-holder rule"
+        );
+    }
+
+    /// Slice 2 (b): the carry — capture-stripped from the hash before
+    /// the router or the session-windows reader evaluates (the ?token=
+    /// pattern applied to the fragment), seeded into the same persisted
+    /// keys this origin would have written itself, the pre-swap stamps
+    /// one-shot for the completion compare, and the legacy federation
+    /// bearer carried across the hop (R-A4 — per-origin storage never
+    /// follows an origin change by itself). R-A2's shared token helper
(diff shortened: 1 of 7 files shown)


lovon-spec:
Queue ejection record: the first merge-group run (30778108550) failed on the **Windows leg only** — `access::iam::tests::load_state_cached_matches_uncached_and_sees_external_writes` (iam.rs:4473), an IAM cache test untouched by this diff (5523 passed / 1 failed; Linux+macOS full suites green in the same group run, and the same bins suite passed on all three PR legs at this head). Ejection signature was CLEAN + MERGEABLE + zero failed head checks + head unchanged → requeued per the standing flake-requeue practice; entry confirmed QUEUED pos 1. If this test ejects a second group run, it's a pattern (Windows mtime/cache-coherence smell) and deserves its own fix PR rather than another requeue.

PR #790, opened 2026-08-05
Windows install-flow hardening: window-vanish fix, cert auto-import, tokened-URL auto-open

lovon-spec:
## What

Owner Windows-e2e finding 2: after the "machine I'm on now" install, the terminal vanished with all output, there were no next steps, and the dashboard looked "broken" without certs. Agenda card `01KZ7ZWKF05Y0HF8BV6R2135Q8`, items 1-9.

**install.ps1** (the bulk):
1. **Window-vanish fix** — `Fail()` used `exit 1` and the run tail `exit $LASTEXITCODE`; under the documented one-liner the script is a scriptblock in the user's own session, so `exit` closed the window and wiped every diagnostic. Failures now pause (`Read-Host`, skipped when input is redirected) and `throw`; the success path never exits the session.
2. **End-of-run success block** — checkout, binary, shortcuts, the tokened dashboard URL, per-mode daemon status, log paths, restart/always-on instructions.
3. **Auto-open after readiness** — polls `intendant ctl dashboard-url` (up to 120s), then `Start-Process` on the tokened loopback owner URL. install.sh gets the twin (backgrounded poll + `open`/`xdg-open` where a GUI opener exists; survives the `exec`).
4. **Cert auto-import** — runs `intendant access setup` (idempotent) and then executes the CurrentUser imports setup previously printed as homework: CA → `Cert:\CurrentUser\Root` (thumbprint pre-check so the consent dialog fires once), `client.p12` + `p12_password` → `Cert:\CurrentUser\My`. Every step soft-fails to the tokened loopback lane.
5. **User PATH + Desktop/Start-menu shortcuts**.
6. **Persistence stated honestly** — the laptop default stays foreground (now in its own console window, so the installer window and one-liner session survive); the success block says plainly that closing that window stops the daemon and nothing restarts it at logon, and names `intendant service install --now` (unelevated = logon task, elevated = at-boot) for always-on.
8. **Edge app-mode** — the shortcuts run a generated launcher (`%LOCALAPPDATA%\Intendant\launch-intendant.ps1`) that re-reads the fresh per-boot token via `ctl dashboard-url` on every click (a static URL would go stale), TCP-probes the port (the token sidecar can outlive a dead daemon across reboots), starts the daemon when nothing listens, and prefers `msedge --app=`; falls back to the default browser.

**Rust + docs** (following commits): platform-aware first-boot hint (no `serve-certs` on Windows), `127.0.0.1` instead of `0.0.0.0` in the printed dashboard URL, and the getting-started.md:744 doc-lag fix (certless clients *do* complete the handshake — client auth is `OptionalCa`; the loopback token lane is the certless owner path).

## Ships when

**install.ps1 / install.sh are release assets** — the canonical copies are stamped and attached per tag, so this fix reaches users at the **next `v*` tag**, not on merge. (The repo copies are the unstamped source.)

## Constraints honored

- install.ps1 stays **pure ASCII** (release.yml strict-decodes it at stamp time; Windows PowerShell 5.1 BOM-less ANSI pin) and PS 5.1-safe (no ternaries/null-coalescing; native-stderr redirection shielded under `$ErrorActionPreference = "Stop"`).
- PSScriptAnalyzer unavailable on this box (no pwsh) — careful line-by-line review instead, per the card.

### DIFF
--- a/docs/src/getting-started.md
+++ b/docs/src/getting-started.md
@@ -46,6 +46,19 @@ stopping to name the command for you to run. Rust and the native build
 dependencies follow the same law via the per-OS setup scripts from the
 cloned tree.
 
+The install ends concretely, not silently: once the daemon reports ready, the
+installer prints the dashboard's tokened loopback owner URL (`intendant ctl
+dashboard-url` reprints it any time; the token rotates each boot) and opens it
+in a browser where a GUI opener exists. On Windows, `install.ps1` additionally
+imports the generated access certificates into the user's certificate stores
+(the root import asks for consent; a declined dialog costs only browser mTLS —
+the tokened URL still works), puts the binaries on the user PATH, and creates
+Desktop/Start-menu shortcuts that start the daemon when needed and open the
+fresh tokened dashboard, preferring an Edge app-mode window. Failures pause on
+screen instead of closing the window, and without `-Service` the daemon runs
+in its own console window — closing that window stops it, and nothing
+restarts it at logon.
+
 For reproducibility, pin the tag instead of `latest`:
 `…/releases/download/v0.1.0/install.sh`. Release assets are immutable, and a
 stamped installer double-checks itself: it prints its release identity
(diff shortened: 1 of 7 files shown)


lovon-spec:
Local battery: workspace clippy -D warnings ✓, `cargo test -p intendant --bins` ✓, lib crates ✓, e2e 47/48 — the one failure (`held_by_rows_ride_the_successor_catalog_and_serve_the_composer_lane`, a successor-boot 180s timeout) re-ran green in isolation (67s); it fell to compile-governor load contention during the full battery, and its machinery (handover WS/catalog over explicit 127.0.0.1) consumes nothing in this diff. cargo fmt --check ✓, install.ps1 strict-ASCII ✓, `sh -n install.sh` ✓, `node --check` on both edited rigs ✓. PSScriptAnalyzer unavailable on this box (no pwsh) — careful line-by-line PS 5.1 review instead, which caught and fixed two latent hazards (dangling-operator continuations; early-pipeline-stop $LASTEXITCODE staleness).

lovon-spec:
Merge-queue ejection 04:47Z: the group's `test (windows-latest)` failed on the known Windows IAM cache flake — `access::iam::tests::load_state_cached_matches_uncached_and_sees_external_writes`, `assertion failed: Arc::ptr_eq(&hit, &cached)` (iam.rs:4473), 5551/5552 passed; unrelated to this diff (no IAM code touched). Re-armed; re-entered the queue at 04:49:02Z (position 2).

PR #885, opened 2026-08-28
docs(design): MCP Control Lane — design research

lovon-spec:
Design research, posted as a **draft on purpose** — this PR is a discussion vehicle, not a landing request. Nothing in it is commissioned.

**What:** `docs/design-mcp-control-lane.md` (following the existing `docs/design-*.md` convention) — a proposal for full control of Intendant by MCP-only agents, locally and over the Internet, with the CLI's context efficiency.

**Thesis:** make the `ctl` grammar the canonical control-plane API and make MCP one of its transports. Three moves: (1) a facade tool profile (~4 meta-tools resolving the ctl verb tree in-process, authorized per resolved operation; ~1–1.5k tokens of schemas vs 17–29k today); (2) an enrolled agent identity for full-fidelity remote control, carried by a stdio sidecar over the existing peer mTLS + relay machinery; (3) a daemon-terminated OAuth 2.1 "agent lease" lane for zero-install clients, reusing the hosted-control doorbell/preset/floor machinery. Connect never terminates MCP and never mints authority.

**Reading order:** §0 thesis → §5 remote lanes → §10 open questions. §1 is the current-state survey (file:line anchors are against `8b390e99`, this PR's base); §9 stages the work M0–M5, of which M0/M1 are pure wins independent of the remote story.

Review here or on the open questions in §10.

### DIFF
--- a/docs/design-mcp-control-lane.md
+++ b/docs/design-mcp-control-lane.md
@@ -1,0 +1,212 @@
+# MCP Control Lane — design research
+
+**Date:** 2026-08-28 · **Status:** design research — a proposal for discussion; nothing here is scheduled or commissioned · **Base:** `main` @ `8b390e99`
+**Prompt:** an original early-Intendant idea, revisited — *"an agent that only speaks MCP can fully control Intendant"*, preferably as a wrapper around the CLI (progressive discovery, context efficiency, skills), including over the Internet, possibly through Connect.
+
+Provenance: a source survey against `8b390e99` plus an external review of the MCP specification and client ecosystem, both on 2026-08-28. File:line anchors below are against `8b390e99`.
+
+---
+
+## 0. Thesis
+
+**Make the `ctl` grammar the canonical control-plane API, and make MCP one of its transports.**
+
+The inversion in the prompt ("MCP wrapping CLI") is not a new subsystem — it is three bounded moves on machinery that already exists:
+
+1. **The facade** — a new MCP tool profile exposing ~4 meta-tools (`run`, `help`, `docs`, `events`) whose grammar *is* the `ctl` verb tree, executed in-process against the same dispatch and the same per-tool IAM gate. Context cost drops from ~17–29k tokens of tool schemas to ~1–1.5k, with everything else discovered lazily through help text and skills — exactly the doctrine `ctl` already encodes.
+2. **A machine identity for full-fidelity remote control** (lane R1) — an enrolled, daemon-minted agent identity (peer-class mTLS cert with an owner-chosen profile ceiling), carried by a thin sidecar because MCP clients cannot present client certificates. Rides the existing Connect relay under `relay_peer_admission`. This is the "fully control over the Internet" lane; every ceremony it needs already ships.
+3. **A zero-install hosted lane** (lane R2) — daemon-terminated OAuth 2.1 on the fleet name through the ciphertext relay, minting a new *agent-lease* credential class that reuses the hosted-control doorbell → trusted-surface-approval → compiled preset/floor/wall machinery. This is what lets claude.ai/ChatGPT/Cursor-class clients connect with nothing installed — and it is capped exactly the way the hosted browser lane is capped, on purpose.
+
+Connect never terminates MCP, never sees plaintext, never mints authority. The two-layer trust stance survives verbatim: `role:none` hosted floor; daemon-minted, trusted-surface-approved leases as the convenience carve-out; full authority only for locally-enrolled identities.
+
+---
+
+## 1. Where we actually are (survey)
+
+### 1.1 The daemon is already an MCP server — an under-shaped one
+
+- `POST /mcp` on the web gateway: **stateless JSON-RPC** — `initialize`, `tools/list`, `tools/call`, notifications acked with 202; `GET`/`DELETE` are **405 by design** (`gateway_routes.rs:2963–2992`, `mcp_gate.rs:415–536,680`). No SSE, no pagination, no resources/prompts on HTTP. `initialize` **hardcodes protocol `2024-11-05`** and capabilities `{"tools":{}}` (`mcp_gate.rs:455–462`).
+- **Two disjoint tool tables**: 13 tools in the rmcp `#[tool_router]` block (`mcp/mod.rs:1570–2340`) + 43 hand-written HTTP-only definitions (`tool_gate.rs:404–757`). Net: **56 advertised / 75 callable / 19 callable-but-advertised-nowhere** (controller-restart family, browser workspaces, `take_display`, `spawn_live_audio`, …). `--mcp` stdio serves **only the 13 router tools** (rmcp `2025-06-18`, with resources+subscribe that HTTP lacks).
+- **Context cost, measured/estimated**: names+descriptions alone = 25,277 B exact; full unfiltered `tools/list` ≈ **95–115 KB ≈ 26–29k tokens** (±25%, schema sizes estimated from source). The `core` profile every supervised backend gets (26 tools) ≈ **60–75 KB ≈ 17–19k tokens**. `agenda_op` alone ≈ 31–41 KB — one tool outweighs most entire MCP servers. The repo already ratified the countermeasure in miniature: `remote_command` was pulled from `core` explicitly for "context rent" (`tool_gate.rs:146–169`, 2026-08-07).
+- **IAM is the solid part**: one principal ladder on `/mcp` (`mcp_http_access_context`, `mcp_gate.rs:759` — peer mTLS → session-scoped token → process token → own/app browser origin → mTLS cert → tokenless loopback + per-boot admission token), one per-tool map `mcp_tool_operation` (`tool_gate.rs:234–357`), enforced at call time and mirrored into `tools/list` filtering so advertised == callable. Pinned invariant: **no MCP tool reaches `AccessManage`, `CredentialsManage`, `Terminal*`, `Filesystem*`, `Settings`** (`tool_gate.rs:1429` et al.).
+- Four ingresses share one dispatcher (`call_tool_by_name_as_caller`, `mcp/mod.rs:543`): gateway `/mcp`, stdio, the dedicated session-MCP loopback listener (session-scoped tokens only, serves only `/mcp` — `listener.rs:49,133`, `mcp_gate.rs:707`), and the dashboard tunnel's `api_mcp_tool_call` (`dashboard_control/api_control.rs:1196`).
+
+### 1.2 `ctl` is already the context-efficient control surface — and already an MCP client
+
+- `src/bin/caller/ctl.rs` (7,953 lines) + `ctl/remote.rs`: hand-rolled parser (no clap), **24 verb families, ~100 leaves**, dispatching as a stateless JSON-RPC client of `/mcp` (`rpc()` at `ctl.rs:5252`). Mapping is shaped, not 1:1 — 27 agenda leaves compress onto 3 tools; ~64 of the 75 tools are fronted; 2 leaves ride a loopback `/api` lane instead; `takeover` is HTTP-only.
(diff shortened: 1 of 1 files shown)


lovon-spec (commented):
Overall, I think this is a strong and unusually well-researched direction. The context-rent diagnosis is convincing; reusing the existing call-time IAM model instead of inventing a blanket “MCP operator” privilege is exactly right; the R1/R2 distinction is honest; and keeping Connect as a ciphertext relay rather than an authority mint is the right invariant.

I would support M0 now. I also support the facade goal, but I would narrow/reframe M1 before commissioning it. My main recommendation is to change the thesis slightly:

> Make a declarative **control-command registry** the canonical API; make `ctl` and MCP two projections of it.

The current `ctl` parser is not itself a safe server-side API boundary. I would treat the points below as design requirements for M1, and treat M3/M4 as separate security reviews rather than simple follow-ons.

## 1. Do not mechanically reuse the current `ctl` grammar inside the daemon

The document’s intended seam—`parse(argv) -> PlannedCall`—is good, but the current CLI grammar mixes pure command planning with local-process behavior:

- `parse_global_args` accepts target/identity selectors (`--url`, `--peer`, `--port`, `--session`, `--managed-context`) and reads environment-derived routing state.
- `dashboard-url` reads the local sidecar/admission token and prints an owner URL containing that token.
- `tools call` accepts an arbitrary MCP tool name and arbitrary JSON. It is still IAM-gated, but it escapes the proposed closed command registry, including into callable-but-unadvertised/future tools.
- `read_json_value` expands `-` from stdin and `@path` from the filesystem.
- output handling can write/copy to a caller-selected local path.
- `ensure_help` terminates the process with `std::process::exit`.
- most importantly, `ctl remote start -- ...` deliberately carries arbitrary command argv plus `--cwd` and `--env` into the `ShellSpawn` operation.

That last point means §6’s current wording (“no env influence, no cwd semantics, no arbitrary binaries”) is too strong. The accurate security claim is: **the facade router performs no shell parsing or ambient process execution, but some explicitly authorized leaves are intentionally open-world capabilities**. `remote start` is still safe to expose under a `ShellSpawn` gate; it just cannot be used as evidence that the whole surface is closed in the stronger sense.

I would define a pure registry/IR along these lines:

```text
CommandSpec {
  stable_command_id,
  cli_aliases,
  argument_schema,
  risk_class,
  policy_target,
  availability_predicate,
  executor,
  result_schema,
}

plan(command_id, structured_args) -> PlannedCall
```

The CLI frontend can then own environment/config discovery, `@file`/stdin expansion, output files, peer selection, formatting, aliases, and exit codes. The facade should accept a stable command id plus structured arguments (or, second-best, a deliberately restricted pure argv parser), and it should omit/reject CLI-only surfaces such as global transport selectors, `dashboard-url`, `tools call`, `@path`, stdin expansion, and output paths.

This retains the token-saving meta-tool design while avoiding stringified JSON, quoting failures, and accidental daemon filesystem/process semantics. `help(command)` can return the selected command’s generated schema lazily. I would also keep help out of `run` entirely—do not send `--help` through executable planning when there is already a dedicated `help` tool.

The executor should return native structured/MCP content, not parse or emulate CLI stdout. The CLI renderer should be downstream of the same result object.

## 2. Make resolved authorization part of the common authorization plumbing

Today the HTTP ingress authorizes the fixed MCP tool name before dispatch:

```text
access.decision(mcp_tool_operation(name))
```

A facade envelope cannot safely be inserted as an ordinary fixed-operation tool. Mapping `run` broadly would prevent low-authority callers from using read-only commands; mapping it weakly makes correctness depend on every ingress remembering a second, inner gate.

I suggest making the registry expose an authorization target such as:

```text
AuthorizationTarget::Fixed(PeerOperation)
AuthorizationTarget::Resolved(PolicyTarget)
```

and have every ingress—gateway MCP, stdio, session-MCP, and dashboard tunnel—use the same resolve-before-authorize path. Resolution must be pure, bounded, and side-effect-free. Unknown command, unknown operation, or unavailable command must fail closed.

`PolicyTarget` should retain more than a `PeerOperation`. For commands that resolve to HTTP/dashboard routes, the operation is only the first-stage floor: hosted control also has exact route/method walls and concrete action/tenant checks. An R2 facade must execute the same composed policy as the native route, not reduce authorization to `preset_allows_operation(...)` alone. Otherwise broad classes such as `AgendaWrite`/`MemoryWrite` can obscure the narrower exclusions the current hosted wall deliberately encodes.

I would generate and pin a negative authorization matrix over **every command × principal/profile/preset**, including the trust-critical subverbs (approval resolution, memory judgment, agenda stamping/definitions/reminder policy, peer use, runtime control, IAM/credentials). That turns “the facade lifts no ceiling” into an executable invariant.

## 3. A single `run` tool collapses client-side safety semantics

MCP annotations are static hints on a tool definition. One `run` definition would represent all of these at once:

- read-only status/search,
- user messaging/task creation,
- destructive display input,
- approval resolution,
- runtime control,
- arbitrary remote command execution.

The server-side resolved IAM check can remain correct, but host policy and UX become all-or-nothing. Per-tool allowlists, confirmation rules, and `readOnlyHint`/`destructiveHint`/`idempotentHint` cannot distinguish the argv. Likewise, a header-only classifier sees `Mcp-Name: run`, not the resolved leaf.

I would keep the compact facade but split it by risk, for example:

- `inspect` / `query`: provably read-only commands;
- `act`: ordinary mutating commands;
- `authorize`: approval/authority-affecting commands;
- plus `help`, `docs`, and `events`.

The exact names are less important than preserving a truthful read-only tool and giving hosts a separately controllable high-risk lane. This is still a tiny schema budget. If a single `run` remains, its annotations should be maximally conservative and M1 acceptance should explicitly validate the resulting confirmation/allowlist behavior in target clients—not just whether the commands execute.

## 4. Specify protocol support as a compatibility matrix, not “echo the requested era”

The protocol section currently makes 2026 support sound mostly like negotiation. It is a distinct wire contract. The current 2026 spec removes the initialize/session model and adds mandatory protocol-shape details including `server/discover`, the method/name HTTP headers, request metadata rules, deterministic list ordering, and cache metadata (`ttlMs` **and `cacheScope`**), with request-scoped SSE behavior.

I would document and test separate lanes:

1. exact legacy versions supported through `initialize` (2024/2025), with honest capabilities;
2. the 2026 stateless path, advertised only when every applicable MUST is implemented;
3. an explicit unsupported-version response rather than best-effort version echoing.

M0 can absolutely improve the existing legacy negotiation first, and M1 can ship on that base. I would not claim 2026 conformance until the full mandatory set lands. Add integration tests using current official SDKs for each advertised era, including header/body mismatches, cache fields/order, notifications, JSON versus SSE responses, and version errors.

The auth registration wording should also be softened: CIMD is the preferred modern path, but the ecosystem still includes preregistered clients and DCR-era compatibility. R2 needs a per-client compatibility table rather than assuming one registration mechanism.

## 5. R1 reuses machinery, but it is still a new trust boundary

The sidecar/mTLS approach is the right sovereign lane. However, an agent operator should not simply become a peer daemon with a different profile label.

The current relay provenance carve-out is keyed on `peer_connection_identity.is_some()` before ordinary route handling, and `peer_daemon` maps to the Peer actor class. Reusing that identity class for agent credentials changes which relay requests enter the authenticated ladder and can also affect tenant-edge semantics. IAM still matters, but this is more than transport reuse.

I would require:

- a distinct principal/credential type such as `agent_operator`;
- an explicit actor classification;
- certificate purpose/audience separation (separate issuer/profile/EKU or an equally strong cryptographic discriminator);
- relay admission scoped to the intended endpoint set, initially `/mcp` only;
- explicit denial on peer-federation, dashboard, access, and administrative routes;
- a cautious default profile, with Approval, ShellSpawn, DisplayInput, and similarly strong operations granted explicitly;
- enrollment, rotation, revocation, lost-key, expiry, and audit semantics;
- a sidecar configuration that resolves a pinned enrolled target, never an arbitrary URL supplied by MCP argv.

I would change “M3 needs no new trust machinery” to “M3 introduces a new principal/credential boundary while reusing the existing transport and enrollment machinery.”

## 6. Make R2 its own security RFC and split M4

R2 is plausible and the doorbell reuse is elegant, but it is the genuinely large security/product surface here. I would split M4 into at least:

1. OAuth AS/RS conformance and credential plumbing;
2. the `agent_lease` bridge into the hosted policy compiler;
3. client-specific pilots and activation.

Before implementation, build a target-client matrix covering protocol era, HTTP transport, authorization-server discovery, client registration mode (pre-registration/CIMD/DCR compatibility), requested scopes, refresh behavior, tool-catalog refresh behavior, output limits, and redirect constraints.

The security design should explicitly cover exact redirect validation, authorization-code binding to client id/redirect URI/PKCE/resource/scope, state and issuer mix-up defense, one-time redemption, access-token audience, revocation, key rotation, multiple-credential ambiguity, rate limits, and audit. If CIMD metadata is fetched by the daemon, that is a public server-side fetch surface: specify SSRF/private-address denial, redirect policy, DNS rebinding defense, response size/time limits, content-type validation, and caching.

For refresh semantics, I recommend separating access-token lifetime from the approved lease envelope: short-lived access tokens plus rotating, replay-detected refresh tokens that can never extend beyond the original trusted-surface-approved lease expiry. Extending beyond that expiry requires a new doorbell. That preserves the ≤24 h authority decision without forcing connectors to reauthorize whenever a short access token expires.

My answer to the ceiling question: **Tasks is a reasonable useful default; Operate should be an explicit, short-TTL step-up with a separate consent display, not the default.** I would initially keep terminal/raw-filesystem capabilities out of R2 even if they remain available to explicitly granted R1 identities. Make `agent_lease -> Unattributed` an explicit classification/invariant rather than depending on a generic unknown-kind fallback, and pin the full forbidden-operation matrix.

## 7. `events` needs object-level authorization, not only `SessionInspect`

A daemon/session event stream can aggregate records with different audiences. Please specify:

- which event classes each principal can see;
- session/tenant projection and redaction before cursoring;
- whether cursors are cryptographically/structurally bound to principal + filter + session scope;
- expiry, retention, gap/compaction behavior, and restart semantics;
- output caps and continuation when a single interval is too large;
- long-poll concurrency/rate limits, cancellation, and backpressure.

A cursor minted under a stronger principal must not be replayable under a weaker one, and a session-scoped caller must never use the daemon-wide cursor namespace to cross session boundaries. These deserve explicit acceptance tests.

## 8. Suggested M1 acceptance gates

In addition to the end-to-end Claude Code scenario, I would require:

- direct typed-tool call and facade call yield the same authorization decision and semantically equivalent result for every registered command;
- every facade command has a stable id, typed schema, policy target, risk class, and availability rule;
- no facade input can select transport/identity, read stdin/files/env, write daemon-local files, exit the process, or invoke an unregistered tool;
- parser/planner fuzz/property tests prove no side effects occur before authorization;
- generated drift tests cover CLI aliases/help, facade discovery, direct tool tables, and policy classification;
- a schema/context report is reproducible in CI (bytes plus a named tokenizer), so the ≤2k-token acceptance target does not remain an estimate;
- at least one schema-eager client and one confirmation/allowlist-heavy client are exercised, because the most likely facade failure is client policy/UX rather than server dispatch;
- text/structured/image results have separate, explicit size/truncation rules.

## Answers to the open questions

1. R2 default: **Tasks**; Operate only explicit short-lived step-up.
2. Terminal/raw-fs: initially R1-only, excluded from R2.
3. Supervised backends: experiment after the external/local facade proves authorization parity and client UX; Codex first is sensible.
4. `--mcp` convergence: yes, but retain a compatibility profile/deprecation window for consumers of the 13-tool surface.
5. Naming: `facade` is clear; I favor risk-separated `inspect`/`act`/`authorize` over one `run`.
6. Refresh: rotating refresh within the original approved lease expiry; re-doorbell to extend authority.
7. Program placement: commission M0 and the registry/planner portion of M1 first; track R1 and especially R2 as independently reviewable security milestones.

Net: strong proposal, and I think the small-surface MCP facade is worth pursuing. The crucial correction is to make a pure declarative command/policy registry—not today’s terminal parser—the shared source of truth, and to preserve risk distinctions that MCP hosts can actually enforce and present to users.

PR #886, opened 2026-08-28
mcp: protocol version negotiation + unknown-profile core fallback

lovon-spec:
First implementation slice of the MCP control-lane program (design: #885, `docs/design-mcp-control-lane.md`) — the M0 protocol-hygiene items that are pure wins regardless of the rest of the program.

**1. `/mcp` protocol version negotiation.** `initialize` hardcoded `2024-11-05` no matter what the client requested. It now negotiates per spec: a requested revision the endpoint implements (`2025-06-18`, `2024-11-05`) is echoed; anything else gets the newest implemented revision and the client decides. `2025-03-26` is deliberately not advertised: that revision makes JSON-RPC batching mandatory, and this endpoint parses single-object bodies only (noted in the new `SUPPORTED_MCP_PROTOCOL_VERSIONS` doc). The stdio transport already negotiates via rmcp and is untouched; the `2024-11-05` string in the codex external-agent test mock is a fixture, not wire code, and stays.

**2. Unknown `tool_profile` values fall back to `core` instead of the full surface.** Flagging this one for review attention because the old behavior was a documented deliberate decision ("typoed third-party URLs should not silently hide tools"). The reconciliation: the full unfiltered list is itself the failure mode the profiles exist to prevent (tens of KB of tool schemas swamping a session's context), profile shaping never gates calls (hidden tools stay callable), and the listing edge now logs the unknown profile name — so a typo stays diagnosable while the standard operating set is still advertised. The old comment's rationale is preserved and answered in the new comment.

Tests: negotiation unit + on-the-wire coverage; unknown-profile ≡ core pinned through the real serving path in both managed-context states; a known-profile-list pin so a future profile arm that forgets the log list fails a test; stale fail-open comments in existing tests updated.

### DIFF
--- a/docs/src/mcp-server.md
+++ b/docs/src/mcp-server.md
@@ -71,8 +71,9 @@ browser-workspace and raw frame tools, which currently have no served
 `tools/list` definitions anywhere and are reachable only by direct call;
 `managed` (alias `managed-context`) advertises `get_status` plus the managed
 rewind/fission set; `full` — or omitting `tool_profile` — keeps the whole
-list, and unknown profile names fail open so a typoed URL does not silently
-hide tools. Profile filtering applies to `tools/list` only —
+list, and unknown profile names fall back to the `core` bootstrap set (the
+daemon logs the unknown name, and hidden tools stay callable, so a typoed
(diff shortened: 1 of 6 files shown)


lovon-spec:
@codex review

PR #888, opened 2026-08-28
mcp: CLI-shaped facade profile (M1 slice 1)

lovon-spec:
M1 slice 1 of the MCP control-lane program (design: #885 `docs/design-mcp-control-lane.md`; M0 landed in #886/#887). This is the facade itself, local-first.

**What ships:** `tool_profile=facade` — five meta-tools replacing dozens of typed schemas. `inspect` / `act` / `authorize` execute one registered command per call as an argv array; `help` renders the command map from the registry; `docs` serves the embedded operating skills. The command registry (`mcp/facade.rs`, 27 starter commands across status/approvals/input/ask/notify/notes/task/agenda/memory/display) is deliberately pure: no filesystem, environment, stdin/`@file` expansion, output paths, or process exit daemon-side — those stay in the `ctl` frontend.

**The authorization law** (per the design review): resolve-before-authorize at every live ingress. Both the HTTP gate and the dashboard tunnel resolve a facade call to its command and authorize the *command's* operation — derived from the underlying tool via `mcp_tool_operation`, declared once — against the caller's principal. Parse failures never dispatch. Wrong-lane calls are redirected by name so host-side per-tool confirmation policy stays truthful. The meta-tool names themselves sit on the restrictive `RuntimeControl` default as a guard for any ingress that skips gate-side resolution.

**Risk lanes:** `inspect` = provably read-only (pinned: every inspect row resolves to a read-class operation), `act` = ordinary mutation, `authorize` = approval/authority class (pinned: `Approval`-op commands ride only here). Hosts can free-list `inspect`+`help`+`docs` and gate `authorize` hard.

**Budget:** the whole serialized facade `tools/list` is test-pinned under 8 KiB (~1.5–2k tokens) versus ~60–75 KB for `core` and ~95–115 KB unfiltered.

**Not in this PR:** the terminal verb family (lands with its PTY accessor plumbing next, per the owner ruling), `events` long-poll (M2), stdio convergence (open question Q4), remaining ctl families (migrate incrementally).

PR #889, opened 2026-08-29
mcp: terminal tool family (M1b)

lovon-spec:
M1b of the MCP control-lane program (design: #885; facade landed in #888): the terminal verb family, per the owner ruling that controlling agents get terminal access.

**Six tools over the existing dashboard PTY pool** — `terminal_list` / `open` / `read` / `write` / `resize` / `close`. Same `TerminalRegistry`, so a shell opened over MCP is attachable from a dashboard terminal tile and vice versa, and the registry's own visibility model applies (root surfaces see everything; scoped principals see their own and shared sessions, derived from the bound IAM principal — a missing principal id owns nothing).

**Structurally honest operations.** Where the dashboard tunnel's `terminal_open` frame decides attach-vs-create statefully inside the handler, here it is two authorities: reads ride `terminal.view`, input/resize/close ride `terminal.write`, and `terminal_open` always carries `shell.spawn` because it creates when absent. No stateful check, no operation broader than the tool.

**Polling, not streaming** — additive plumbing on `terminal.rs`: a monotonic total-bytes-written cursor over the scrollback ring with `read_since` (reports `gap: true` when the cursor fell off the 256 KiB window — the polling analogue of the push lane's dropped-output marker), sharing `attach`'s critical section; a retained exit status so a poller that missed the death still learns it; a geometry accessor; and `list_visible`, which skips in-flight spawn reservations like every other read.

**Facade rows:** `terminal list`/`read` on `inspect`; `resize`/`close` on `act`; `open`/`write` on `authorize` — writing into a live shell is running commands, the same class as spawning one. Terminal tools stay out of the scoped advertisement profiles (the `remote_command` context-rent precedent); they appear on full/unprofiled listings and through the facade.

Hosted note: nothing in this PR touches the hosted walls — `/mcp` remains excluded from the hosted-lease route set, and the R2 question ("maybe for R2", Operate-preset terminal) stays a policy decision for the M4 security review, exactly as ruled.

### DIFF
--- a/docs/src/mcp-server.md
+++ b/docs/src/mcp-server.md
@@ -405,7 +405,11 @@ authorities here). Output is polled with a monotonic cursor over the
 marker — and the exit status is retained so a poller that missed the
 death still learns it. Visibility is the registry's own model: root
 surfaces see every session, scoped principals see their own and shared
-ones. None of these ride the scoped profiles; they appear on
+ones. A shell spawned for a scoped principal is OS-sandboxed to the
+grant's filesystem scope and — independently, even when the grant
+carries no scope — never inherits the daemon's process environment (the
+daemon env holds provider API keys; the child env is cleared and
+rebuilt secret-free). None of these ride the scoped profiles; they appear on
 full/unprofiled listings and through the facade's `terminal` commands
 (`open` and `write` on the `authorize` lane — writing into a live shell
 is running commands).
@@ -415,7 +419,7 @@ is running commands).
 | `terminal_list` | Visible sessions: id, liveness, sharing, geometry, retained exit status. | — |
 | `terminal_open` | Open or create a shell PTY (shell-spawn class); returns the id, whether it was created, geometry, and the starting read cursor. | `terminal_id?`, `cols?`, `rows?`, `shared?` |
 | `terminal_read` | Cursor-paged output read with gap reporting, liveness, and exit status. | `terminal_id`, `cursor?`, `max_bytes?` |
-| `terminal_write` | Write to a live shell's stdin (appends a newline by default; `enter: false` for raw keystrokes). | `terminal_id`, `input`, `enter?` |
+| `terminal_write` | Write to a live shell's stdin (appends Enter — a carriage return — by default; `enter: false` for raw keystrokes). | `terminal_id`, `input`, `enter?` |
 | `terminal_resize` | Resize the PTY. | `terminal_id`, `cols`, `rows` |
 | `terminal_close` | Close the session. | `terminal_id` |
 
--- a/docs/src/mcp-server.md
+++ b/docs/src/mcp-server.md
@@ -410,7 +410,8 @@ grant's filesystem scope and — independently, even when the grant
 carries no scope — never inherits the daemon's process environment (the
 daemon env holds provider API keys; the child env is cleared and
 rebuilt secret-free, and the shell starts profile-less so rc files
-can't repopulate it). The sandbox is fixed at spawn, so a caller-owned
+can't repopulate it — a shell with no known profile-suppression mode is
+substituted by profile-less bash). The sandbox is fixed at spawn, so a caller-owned
 session whose grant scope has since changed is refused as stale by
 open-attach, reads, writes, and resizes — close it and reopen to get a
 shell under the current scope. None of these ride the scoped profiles; they appear on
--- a/docs/src/mcp-server.md
+++ b/docs/src/mcp-server.md
@@ -409,7 +409,8 @@ ones. A shell spawned for a scoped principal is OS-sandboxed to the
 grant's filesystem scope and — independently, even when the grant
 carries no scope — never inherits the daemon's process environment (the
 daemon env holds provider API keys; the child env is cleared and
-rebuilt secret-free). None of these ride the scoped profiles; they appear on
+rebuilt secret-free, and the shell starts profile-less so rc files
+can't repopulate it). None of these ride the scoped profiles; they appear on
 full/unprofiled listings and through the facade's `terminal` commands
 (`open` and `write` on the `authorize` lane — writing into a live shell
 is running commands).
--- a/docs/src/mcp-server.md
+++ b/docs/src/mcp-server.md
@@ -410,7 +410,10 @@ grant's filesystem scope and — independently, even when the grant
 carries no scope — never inherits the daemon's process environment (the
 daemon env holds provider API keys; the child env is cleared and
 rebuilt secret-free, and the shell starts profile-less so rc files
-can't repopulate it). None of these ride the scoped profiles; they appear on
+can't repopulate it). The sandbox is fixed at spawn, so a caller-owned
+session whose grant scope has since changed is refused as stale by
+open-attach, reads, writes, and resizes — close it and reopen to get a
+shell under the current scope. None of these ride the scoped profiles; they appear on
 full/unprofiled listings and through the facade's `terminal` commands
 (`open` and `write` on the `authorize` lane — writing into a live shell
 is running commands).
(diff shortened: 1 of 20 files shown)


lovon-spec:
@codex review

lovon-spec:
@codex review

lovon-spec:
@codex review

Round-3 fixes are up (head 798f1157):

- **tools_terminal.rs:214 (P1, scoped-shell env)** — environment secrecy is now derived from the spawning ACTOR at the registry spawn choke point (`PtySession::spawn`), decoupled from filesystem sandboxing: every principal-owned spawn — every `ToolCallerTrust::Scoped` caller, with or without an fs scope, and the dashboard tunnel / WS lanes' scope-less principal grants too — gets a cleared environment rebuilt from the secret-free allowlist (real home kept; the shell is only secret-free, not fs-sandboxed). Root-lane shells (the owner's own surfaces) inherit as before. The scoped env builders are refactored into home-parameterized secret-free cores, and a new e2e test plants a canary in the daemon env and proves a scope-less principal shell cannot see it.
- **tools_terminal.rs:314 (P1, ConPTY Enter)** — `enter` now appends CR, the byte the Enter key sends and the only line-submit ConPTY recognizes (Unix ICRNL maps it to NL); the tool description, facade flag help, and docs updated to match.
- **tools_terminal.rs:130 (P2, invalid bytes before a split tail)** — `utf8_page_len` now scans past invalid sequences (resuming after each `error_len()`) and still holds back a trailing incomplete sequence, so a binary page ending in a split character no longer destroys it; the mixed cases (`[0xff, a, b, 0xc3]`, `[0xff, 0xc3, 0xa9]`, lone `0xff`) are pinned in tests.

On **tools_terminal.rs:125 (P2, undersized UTF-8 prefixes)** — I believe this one is stale: it restates the round-2 finding against the pre-e02516d3 code. Since e02516d3, `max_bytes` clamps to a floor of 4 (the longest UTF-8 sequence), so `max_bytes: 1` never reaches `read_since`; and when the ring itself holds only a partial sequence, the incomplete prefix is HELD, not consumed — `utf8_page_len` returns `valid_up_to()` (0 for a bare prefix) on the `error_len() == None` path, the cursor rewinds to the boundary, and the next read re-delivers the sequence whole. `utf8_page_len_holds_back_incomplete_tails` pins exactly this (`&s[2..3]` → 0). With this round's :130 fix, no path in the function consumes an incomplete sequence at all — the only bytes ever consumed past a replacement are genuinely invalid ones. If you see a remaining input that consumes a completable prefix on head 798f1157, please spell out the byte sequence and cursor state.

lovon-spec:
@codex review

Round-4 fixes are up (head 2d362714) — all three were real:

- **profile-less principal shells (P1)** — principal-owned spawns now start with the rc-skip argv family the scoped path already used (`secret_free_shell_args`, renamed from `scoped_shell_args`, plus a Windows twin: `-NoLogo -NoProfile` for powershell/pwsh, `/d` for cmd), so rc files can't repopulate the cleared environment; the owner's root-lane shells keep the login-style startup. Args pinned per shell in tests on both platforms.
- **dead sessions release split tails (P2)** — `terminal_read` now delivers the whole page once the shell has exited (`page_keep_len`: hold-back only while alive), so a truncated final sequence can't park the cursor forever; the Windows exit-flag-vs-final-drain race is documented as a one-character decay against a permanently wedged cursor. Pinned in tests.
- **canary restore (P2)** — the e2e env test restores the canary's prior state via a drop guard on every exit path, inside the env-lock window.

lovon-spec:
@codex review

Round-5 fix is up (head e7f98d75) — the finding was real:

- **peer filesystem scope over /mcp (P1)** — `RequestAuthority` now carries a `peer_filesystem` field, populated exactly when `http_access_context` mints the authority from a `PeerConnectionIdentity` (`Some(identity.filesystem)`), and `fs_scope()` returns it ahead of the IAM-derived paths — a peer is never owner-class `None`/unrestricted, matching the dashboard tunnel's `Peer` grant arm. Pinned at both levels: a unit test that a peer-minted authority's `fs_scope()` is the identity policy, and an ingress test that the `http_access_context` peer arm carries it through. Every other authority construction site states `peer_filesystem: None` explicitly.

lovon-spec:
@codex security review

Final security pass requested on the finished shape (head e7f98d75): the security row last ran on the initial head, before the env-secrecy derivation, profile-less principal shells, and the peer filesystem-scope carry all landed.

lovon-spec:
@codex security review

Security-round fix is up (head 0654467f) — the finding was real:

- **stale sandbox after a scope change (P1)** — the session now records the filesystem scope it was spawned under (`PtySession::spawn_scope`), and every MCP access to a caller-owned session — `terminal_open`'s attach path, `terminal_read`, `terminal_write`, `terminal_resize` — compares it against the caller's CURRENT grant scope stated by the ingress gate. Any mismatch (narrowed, broadened, gained, or lost scope) refuses with close-and-reopen as the remedy; `terminal_close` always works so the stale shell can be put down, and a fresh spawn is by definition under the current scope. Shared sessions another principal owns ride that owner's authority by design, and root surfaces are never scope-bound — the staleness predicate is pinned across all of those cases, and the spawn-scope recording is asserted in both spawn e2e tests.

lovon-spec:
@codex security review

Security round-2 fix is up (head 6ef189ab) — the finding was real:

- **unsuppressible startup profiles (P1)** — principal-owned unsandboxed spawns now go through `principal_shell_program`: a shell with a known suppression mode (zsh/bash/fish; powershell/pwsh/cmd on Windows) runs as itself with its rc-skip argv, and anything else — tcsh reading `~/.tcshrc` on every start being the canonical case — is substituted by bash in `--noprofile --norc` mode (profile-less PowerShell on Windows) instead of being trusted. The empty-argv fallback in `secret_free_shell_args` is now documented as safe only under a filesystem sandbox (where rc files are unreadable), and the substitution is pinned per shell on both platforms.

lovon-spec:
@codex security review

Security round-3 fix is up (head 41e541b0) — the finding was real, and the macOS profile carried the same class, so both platforms are fixed:

- **shared same-account trees (P1)** — Linux: the read baseline enumerates ownerless device nodes instead of bare `/dev` (Landlock is additive-only, so /dev-minus-PTYs cannot be expressed) and the write baseline is `/dev/null` + `/dev/tty` only; `/dev/pts`, `/dev/shm`, `/tmp`, `/var/tmp` are gone. macOS: `/dev/ttys*` is denied last-match-wins for read and write, and the `/private/tmp`, `/private/var/tmp`, and daemon-TMPDIR write allows are gone. Both: each scoped spawn creates a fresh 0700 session-private scratch directory, grants exactly that, points `TMPDIR` at it, and removes it when the session drops (spawn-failure paths clean up too). The shell's own terminal rides its inherited descriptors plus `/dev/tty` (controlling-terminal only); nested PTY allocation inside a scoped shell is deliberately unavailable. Pinned: Linux baseline exclusions (runs on the CI Linux leg), Seatbelt scratch/ttys-deny/no-shared-tmp assertions, TMPDIR-is-scratch in the scoped env, and the live macOS sandbox e2e still passes — a real zsh boots and operates under the narrowed profile.

One consciously retained read: `/run` stays in the Linux read baseline (the `/etc/resolv.conf` → `/run/systemd/resolve` chain and NSS need it). Files under `/run/user/<uid>` are same-account-readable in principle — flagging it here for your judgment; unix-socket connects aren't Landlock-fs-gated either way, so the residual exposure is plain files under `/run/user`.

lovon-spec:
@codex security review

Security round-4 fix is up (head 34272282) — the /run ruling implemented:

- **narrow the Linux /run read baseline (P1)** — bare `/run` is gone; only the resolver backends `/etc/resolv.conf` can point at are enumerated (`/run/systemd/resolve`, `/run/resolvconf`, `/run/NetworkManager`). The Landlock applier already skips paths that don't exist on a given distro, so the enumeration is portable, and the baseline pin test now asserts bare `/run` is excluded while the resolver backend stays.

(The same commit repairs the CI clippy break the round-3 commit shipped — a doubled doc comment on the scoped-command builder whose lines parsed as unindented list continuations, plus a single-element loop only macOS clippy sees; the Windows line of that doc also now states the restricted-token reality instead of the stale "refused".)

PR #890, opened 2026-08-29
mcp: the events verb - cursor long-poll over the daemon lifecycle stream

lovon-spec:
M2 of the MCP control lane (docs/design-mcp-control-lane.md): an
MCP-only agent can now WAIT for things to happen instead of polling N
read commands — the transport-agnostic answer to follow/attach that ctl
never had either.

The daemon-wide event ring (event_ring.rs) is a bounded (1024 entries /
2 MiB, drop-oldest) cursor-addressed window fed once per process from
the bus broadcast lane, next to the other mode-independent bus folds.
Its ingest allowlist is the security boundary: only the session,
approval, and task lifecycle families ride — exactly the content the
session.inspect-gated read tools already serve — so the verb is push
semantics over existing authority, never new authority. The streaming
floods (model deltas, agent output, context snapshots) and the families
gated by other operations (agenda, memory, display, credential,
presence) are never ingested. Loss is visible twice over: a lagged bus
receiver records a synthetic event_gap entry, and a cursor that fell
off the window reports gap: true.

The events tool (sixth facade meta-tool, session.inspect at the gate
and in advertisement) long-polls with the remote-wait chunk contract:
wait_s clamps to 60, the Notify wakeup is armed before each read so a
push cannot be missed, and a lapsed wait does one final read before
reporting quiet (the ask-tool ledger discipline). Cursors are opaque
"{epoch}.{seq}.{tag}" strings: the epoch is the ring's per-boot
identity, so a cursor from a previous boot fails loudly instead of
silently reading wrong positions, and the tag binds the cursor to the
minting principal (the design-review amendment) — a cursor minted under
another principal is refused. since omitted starts at NOW (no backlog
replay); filter keeps named events but never suppresses the event_gap
marker.

The ring travels SessionListeners → spawn_mode_web_gateway → McpState;
bare stdio --mcp has no ring and the verb answers unavailable there
(stdio convergence is a separate open question).

### DIFF
--- a/docs/src/mcp-server.md
+++ b/docs/src/mcp-server.md
@@ -70,7 +70,7 @@ served display/CU and shared-view set — its allowlist also names the
 browser-workspace and raw frame tools, which currently have no served
 `tools/list` definitions anywhere and are reachable only by direct call;
 `managed` (alias `managed-context`) advertises `get_status` plus the managed
-rewind/fission set; `facade` advertises only the five meta-tools of the
+rewind/fission set; `facade` advertises only the six meta-tools of the
 CLI-shaped facade (next section); `full` — or omitting `tool_profile` —
 keeps the whole
 list, and unknown profile names fall back to the `core` bootstrap set (the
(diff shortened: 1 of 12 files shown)


lovon-spec:
@codex review

Round-1 fixes are up (head 4bc78c56) — both real:

- **filtered-poll stall (P2)** — the loop now drains the buffered window to the ring's head before considering any wait (nothing already buffered fires the notify), which also makes wait_s=0 scan the full window; pinned with a deterministic 250-filtered-entries-then-one-match test.
- **future cursors (P2)** — cursor validation is consolidated (shape, epoch, principal binding, and position: seq must not exceed current_seq) with every refusal branch pinned, and the ring's gap arithmetic now saturates so out-of-range input can never overflow-panic inside the lock and poison the mutex, even if a future caller skips validation.

lovon-spec:
@codex security review

Final security pass on the finished shape (head 4bc78c56): the security row completed on the initial head; since then the round-1 fixes landed — consolidated cursor validation (shape/epoch/principal/position) and the buffered-window drain. The security-relevant design is unchanged: ingest allowlist = session/approval/task lifecycle only, principal-bound cursors, per-boot epoch, saturating ring arithmetic.

lovon-spec:
@codex security review

Security-round fix is up (head 9c27b778) — the finding was real:

- **session-scoped delivery (P1)** — event delivery now mirrors `get_pending_approval_scoped`'s discipline exactly, via the same `McpToolScope::from_actor` derivation: an agent-session caller (session-bound MCP token) receives only entries whose session matches its gate-bound id; sessionless events fail closed for it; a scope whose gate bound no id sees nothing; the `event_gap` loss marker stays visible (no content beyond "events were missed"). Other caller classes keep the daemon-wide view their `session.inspect` grant already reads, and cursor positions stay global — coarse volume metadata, not content (flagging that deliberately: per-principal sequence spaces would be the escalation if you judge cursor arithmetic sensitive at this operation). Ring entries now carry their tag and owning session extracted once at ingest, and the scoping is pinned at the predicate level and end-to-end through the tool.

PR #891, opened 2026-08-29
mcp: the facade registry reaches full ctl fidelity; ctl gains the events verb

lovon-spec:
M2 slice 2 of the MCP control lane. The command registry grows from 37
to 106 rows and now covers every ctl family that maps to a single MCP
tool call: computer-use actions, the full display family (create,
frames, read-frame, take/release, user-display grant/revoke/request),
browser workspaces, the shared-view family, settings, remote compute
(start/status/wait/cancel — all on the authorize lane: the tool carries
one shell-spawn operation, and lane never overrides op), the whole
agenda write/effect family (patch, lifecycle verbs, graph edges, refs,
attest, schedule/stamp/approve/revoke/withdraw/start), memory owner
curation (accept/dispute/retire/supersede via verdict seeds), the
controller family, live audio spawn, peers (list/message/delegate), and
fission claim. Deliberately still excluded: dashboard-url (local-only),
tools call (variable-op escape hatch), and the loopback /api lane
(takeover, agenda ops/occurrences).

New vocabulary: ValueKind::Json — structured parameters (cu action
arrays, shared-view regions, agenda patches/recurrences/binding-refs,
remote argv/env, peer context, audio response schemas) are passed as
literal JSON and parsed at plan time; malformed JSON is a plan error
that never dispatches. The CLI-side @file/stdin expansions and digest
computation stay client-side as ever.

One advertisement pin sharpened: view-class operations legitimately
carry act rows (the shared-view verbs mutate what the user watches and
have always ridden display.view), so a display.view principal now sees
the act lane; a principal holding only the pure-read ops still does
not. Browser list/providers ride inspect (they are stats reads — ops
verified against the gate map, not the survey's coarse note).

ctl side: `intendant ctl events` — the events verb driven with the
remote-wait chunking idiom (client budget split into ≤60s server
polls). One-shot by default (returns on the first delivered batch);
--follow streams until the --for budget ends; NDJSON on stdout with
the resumable cursor and gap warnings on stderr so pipelines see only
events.


lovon-spec:
@codex review

Round-1 fixes are up (head b97c2625) — all three real:

- **request-occurrence lane (P1)** — moved to the authorize lane beside approve/revoke/start, help text names it authority-class.
- **negative booleans (P2)** — default-true fields are now flippable via Json-kind flags carrying literal `false`: `agenda start --interactive false` (the autonomous goal-run), plus the same-class gaps `controller halt --persistent false` and `remote start --require-clean false`. Parity-pinned as real JSON booleans.
- **final-second poll (P2)** — `ctl events` rounds a positive remaining budget up before the 60s clamp, so `--for 1` makes its one poll.

lovon-spec:
@codex review

Round-2 fix is up (head e46ef0ad) — real, same class as round 1:

- **direct task-start override (P2)** — `task start --orchestrate` is now Json-kind tri-state (`true` forces orchestration, `false` forces a direct session, omitted keeps automatic selection), parity-pinned including the omission case.

lovon-spec:
@codex review

Round-3 fixes are up (head 3236b0c0) — both real:

- **task-start display context (P2)** — the row now carries `--frame` (repeatable → `reference_frame_ids`) and `--display-target`, parity-pinned, so CU-context tasks route to the ephemeral runner instead of the plain loop.
- **NDJSON followed envelopes (P2)** — `--json` prints one compact envelope per line; `--follow` streams stay line-oriented.

lovon-spec:
@codex review

Round-4 fixes are up (head 14c91e5a) — all three real parity gaps, now closed and pinned:

- **rewind handoff lists (P2)** — `context rewind` carries repeatable `--discard`/`--artifact`/`--next-step` (→ `discard`/`artifacts`/`next_steps`).
- **uncapped accessibility reads (P2)** — `cu elements --full-values`.
- **memory project provenance (P2)** — `memory propose --project`.

lovon-spec:
@codex review

Round-5 fix is up (head 35664841) — real, the parity class again:

- **blocking-ask vocabulary (P2)** — the `ask` row now carries the whole decision contract: `--questions` (multi-question JSON form; positional optional), `--previews`, `--pick-min`/`--pick-max`/`--multi`, `--free-text false`, `--consequence`, `--wait`, `--expires`, `--park`, `--session`. Parity-pinned including the park short form. (`--schema` stays CLI-side — it is the @file reader whose facade twin is the `--questions` literal.)

lovon-spec:
@codex review

Round-6 fixes are up (head c24979bb) — both real:

- **flat fields with --questions (P2)** — the planner now rejects the combination at plan time with an error naming the per-question form (the daemon would silently ignore the flat twins). The round-5 test is reworked onto the honest forms plus the rejection pin; call-level fields (`--wait`, `--expires`, `--park`, `--session`) stay flat and travel.
- **target-session routing (P2)** — `notify --session` and `session note --session` carry `session_id`, parity-pinned.

lovon-spec:
@codex review

Round-7 fix is up (head a846960b) — real: `agenda add`'s seed now defaults `kind` to `task`, matching ctl's default for identical usage; help names the default and the seeded-args pin asserts it.

lovon-spec:
@codex review

Round-8 fixes are up (head 9b8b87da) — all three real:

- **latest-frame default (P2)** — `display read-frame` plans without an id (seeds `latest`; positional overrides).
- **acquire holder default (P2)** — `browser acquire WS` plans with `facade` holder provenance (the planner is pure and cannot know the caller's session; the positional overrides for session-attributed leases).
- **sole-blocker clear (P2)** — `agenda unblock ID` seeds the empty blocker prefix the daemon resolves uniquely.

All pinned with default and override forms.

lovon-spec:
@codex review

Round-9 fixes are up (head 528d0e17) — both real:

- **bare agenda list (P2)** — seeds `status: open` (the useful default), `--all` lifts the seed, explicit `--status` overrides; parity-pinned. The residual divergence is named honestly in the row help: ctl's bare list additionally folds in unconsumed answers as a client-side render, which a pure single-call planner cannot replicate — if you judge that union belongs server-side, that is an `agenda_list` tool change (a `scope: working` param) I would take as a follow-up rather than smuggle into the registry PR.
- **one-shot gap (P2)** — `ctl events` treats `gap: true` as terminal in one-shot mode; `--follow` keeps streaming from the advanced cursor.

lovon-spec:
@codex review

Round-10 fixes are up (head 8909c7c2):

- **lease holder collision (P2, real — good catch)** — the seed is now a `__caller` identity sentinel the dispatcher substitutes with the gate-bound session id (else principal, else the facade label), so two facade sessions can never collide as one holder. Substitution happens after gate resolution and only touches argument values, never the authorized operation. Pinned at the planner (sentinel travels) and the substitution.
- **agenda ask plain form (P2, real)** — the row now mirrors ctl's own split: plain TEXT → op add/kind question; `--questions` → the structured op:ask form; mixing or omitting both is a plan error. Pinned all four ways.
- **note images (P2, real)** — `session note --images` takes literal JSON attachment objects (media_type + base64 data); no caller paths.
- **agenda ops/occurrences (documented boundary, not a gap)** — these ride the loopback-token `/api` lane and refuse the peer and supervised-session lanes by design; the facade deliberately does not tunnel through that trust boundary. The docs now name all three exclusions explicitly, narrowing the full-grammar claim as you suggested.

lovon-spec:
@codex review

Round-11 fixes are up (head 5e239f98):

- **atomic park-time refs (P2, real)** — `agenda add --refs` and `agenda ask --refs` take literal ref-object arrays that land in the same append as the park; parity-pinned on both rows.
- **derived list views (documented exclusion, pinned)** — `--blocked`/`--frontier`/`--under` and the bare-list answered-union are ctl client-side renders over the same `agenda list` read (blockers, dependency edges, placement, triage watermark); a pure single-call planner cannot derive them. The row help and the docs now say "fetch and derive", and a pin asserts each render flag fails planning — the exclusion is intent, not omission. If those views should become server-side (an `agenda_list` scope param), that is the same follow-up as the round-9 working-set note.

lovon-spec:
@codex review

Round-12 fixes are up (head 01d407e2):

- **empty one-shot polls (P2, real)** — non-follow `--json` suppresses empty non-gap chunks and emits exactly one meaningful record (delivered batch, gap envelope, or one final timeout envelope); `--follow` remains the print-every-chunk batch stream.
- **approve review form (P2, resolved by op purity)** — ctl's digestless form is a client-side render over the item read; the facade ceremony is `inspect agenda show ID` (serves `effects[].manifest` + digest) then `authorize agenda approve ID DIGEST` — a read must not ride the authorize row's AgendaWrite op. `--digest` is now accepted as the ctl spelling, and the row help walks the two-step.
- **takeover (documented exclusion)** — named as the fourth deliberate exclusion: the daemon-drain act on the loopback lane, owner/local trust class.

lovon-spec:
@codex review

Round-13 fix is up (head 9cdd65ee) — real: the gap-terminal exit no longer falls through to the `nothing arrived` timeout diagnostic; that message (and the quiet `--json` final envelope) now fire only when the budget truly lapsed quiet.

lovon-spec:
@codex review

Round-14 fixes are up (head 0e6a9f59) — all three real, one class: ctl's flag spellings now alias the positional forms (`context rewind --item-id/--reason/--primer`, `cu actions --actions`, `controller schedule --controller-id/--goal`), each filling the same json key so either spelling satisfies the requirement. Parity-pinned across all four alias families including round-12's `--digest`.

lovon-spec:
@codex review

Round-15 fixes are up (head 2a029f3b) — all six real, one class, swept systematically this time:

- `remote start [flags] -- CMD ARGS…` plans verbatim: a new rest-list positional keeps word boundaries as a JSON argv array; `--env KEY=VALUE` (repeatable) folds into the env object; `--allow-dirty` flips `require_clean`; `--revision`/`--timeout` alias the ctl spellings.
- `browser create URL` positional + `--peer`/`--session` ctl spellings.
- `controller complete --restart-id/--token/--summary`.
- `context inspect --item-id`.
- `display request --reason`.
- `audio spawn --args '{...}'` dispatches the whole tool object (decomposed flags win on conflicts).

All pinned in one ctl-primary-grammars parity test.

lovon-spec:
@codex review

Round-16 fixes are up (head 8d57eee0) — both real, the alias class: `context claim-fission --group-id/--branch-session-id/--expected-canonical-session-id` and `controller halt --one-shot` (→ `persistent: false` via the negative-flag transform). Parity-pinned.

lovon-spec:
@codex review

Round-17 fixes are up (head d28cc80e) — all three real: `browser release --holder` alias, `shared focus --region x,y,width,height` (csv → region object, four values enforced, one region only), and `peer task --context` now follows ctl's JSON|TEXT contract via a lenient value kind. Parity-pinned.

lovon-spec:
@codex review

Round-18 fixes are up (head 1d42e92d) — all three real: `shared focus clear` registers as a three-segment path (longest-match resolution), `agenda schedule --goal/--at` (epoch ms; ctl's human `--at` vocabulary remains a client-side conversion, named in help), and `memory supersede --with`. Parity-pinned.

lovon-spec:
@codex review

Round-19 fixes are up (head ab3341cf) — six findings answered as one sweep:

- **prefix-resolving write intake (the semantic one)** — `apply_command` now rewrites every item-id field (primary, target, parent, place-under) to the exact resolved id before any op is appended, so durable ops and folds carry exact ids (the read path's `resolve_prefix` discipline, the ClearBlocker precedent). Ambiguity refuses by name; the id-field map is exhaustive over the command set. Store-pinned including the ambiguous and unknown branches.
- **task start** uses ctl's bare `--orchestrate`/`--direct` pair (both = error) plus `--task`/`--session-id`.
- **agenda patch** takes the ctl flags with explicit clears (`--clear-tags` → `[]`, `--clear-due` → `null`) alongside PATCH JSON.
- **relies-on/relates/ref** register with ctl's single-verb `--remove` spellings (relates refuses `--kind` with `--remove` in ctl's own words; ref's remove drops add-only fields); `place --under`.
- **agenda schedule** carries the full vocabulary: recurrence flags, `--on-unblock`, `--project`, and the common launch pins as dotted `agent_config` fields; stamp and attest aliases too.

Human time/duration vocabularies (`--at "tomorrow 9am"`, `--every "1d"`) remain ctl client-side conversions, named in help — the facade takes epoch/duration ms. All parity-pinned.

lovon-spec:
@codex review

Round-20 fixes are up (head 31870343) — all three real:

- **trigger grammar (P2)** — `--on-item-match KIND:TAG[,TAG…]` parses with ctl's wording; a trigger-only schedule's omitted instant means armed-on-approval via a `__now` sentinel the dispatcher fills with epoch ms (the pure planner reads no clock — the caller-identity discipline); goal-required, cadence-vs-trigger, and no-trigger-no-instant refuse at plan time.
- **stamp launch pins (P2)** — the same dotted `agent_config.*` flags as schedule; the dial knobs are a nested DialConfig and deliberately stay on `--agent-config` JSON.
- **one-command unplace (P2)** — `agenda place ID --remove` sends the empty parent the daemon resolves to the current placement (the ClearBlocker sole-resolve idiom; durable ops carry the exact parent).

All parity-pinned including the refusal branches.

lovon-spec:
@codex review

lovon-spec:
@codex review

lovon-spec:
@codex review

lovon-spec:
@codex review

lovon-spec:
@codex review

lovon-spec:
@codex review

lovon-spec:
@codex review

lovon-spec:
@codex review

lovon-spec:
@codex review

lovon-spec:
@codex review

lovon-spec:
@codex review

lovon-spec:
@codex review

lovon-spec:
@codex review

lovon-spec:
@codex review

lovon-spec:
@codex review

lovon-spec:
@codex review

lovon-spec:
@codex review

lovon-spec:
@codex review

lovon-spec:
@codex review

lovon-spec:
@codex review

lovon-spec:
@codex security review

lovon-spec:
@codex security review

lovon-spec:
@codex security review

lovon-spec:
@codex security review

lovon-spec:
@codex review

lovon-spec:
@codex review

lovon-spec:
Landing note: the code-review loop closed clean (round 43, head 8da5250f). Codex security review ran twice on this PR — each pass found one issue (pre-auth region-CSV allocation; gate-side value parsing), both fixed — but a third security pass over the final two hardening commits could not run: the security-review quota is exhausted (two 'usage limits' responses, ~40 min apart). Those two commits are the gate-split narrowing (the gate now parses strictly less than before) and the per-category dial-approvals merge, both covered by clean code reviews. Queueing on that basis; a follow-up security pass can run once the quota resets.

lovon-spec:
@codex review

PR #893, opened 2026-08-30
Make Xvfb allocation fail closed

lovon-spec:
## Summary

- treat every existing or unreadable X lock/socket entry as occupied and fail cleanly on exhaustion
- keep Xvfb ownership in an exact-child guard with nonblocking Drop and bounded async teardown
- report pre-lifecycle creation failures without a fake display id or teardown side effects
- apply the same allocator semantics to dashboard, cloud-worker, provider, and debug paths
- preserve direct/wire peer event parity

## Verification

- cargo fmt --all -- --check
- targeted intendant-platform and caller tests
- host and Linux-target clippy/check with warnings denied
- two independent blind reviews of the exact diff
- no live Xvfb launch and no process signals during validation

## Deliberately deferred

This slice does not claim atomic scan-to-spawn reservation, process-scoped DISPLAY handling, Xauthority or -nolisten tcp hardening, create-request correlation, or a live Xvfb exercise. Those remain separate follow-up work; consumers should keep live capture disabled until that boundary is completed.

### DIFF
--- a/crates/intendant-platform/src/platform.rs
+++ b/crates/intendant-platform/src/platform.rs
@@ -140,6 +140,27 @@ pub fn request_graceful_terminate(pid: u32) -> bool {
     }
 }
 
+/// Create a FIFO for Unix-only lock-file regression tests. Keeping this tiny
+/// wrapper here preserves `platform.rs` as the crate's documented unsafe
+/// island; production code never calls it.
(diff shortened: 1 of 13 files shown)


lovon-spec:
@codex review exact head db6cd9c9fc439ddafe4d7160f2e2516d3bad5d2e

PR #892, opened 2026-08-30
mcp: per-request SSE for held-POST verbs (M2c)

lovon-spec:
M2's remaining transport slice per the MCP Control Lane design (~/mcp-control-next.md): a `POST /mcp` `tools/call` whose client accepts `text/event-stream` and whose named (or facade-resolved) tool is a held-POST verb (`ask_user`, `request_user_display`) answers as a per-request SSE stream — 15 s keepalives (spec-shaped `notifications/progress` when the request carries `_meta.progressToken`) defeat client/proxy idle timeouts on waits up to 900 s, then the final JSON-RPC response closes the stream. Everything else keeps the plain-JSON leg; `GET /mcp` stays 405 (the 2026-07-28 stateless shape). Authorization unchanged: the per-tool decision runs inside dispatch; SSE mode selection is argv-only (values never parsed pre-auth, the #891 discipline).

PR #894, opened 2026-08-31
mcp: bound the stream-mode facade resolution to the argv prefix

lovon-spec:
Post-merge security pass on #892 (CLI, security-focused, max effort)
verdict: authorization and SSE framing safe; one P2 - the SSE plan's
facade resolution fully materialized caller-controlled argv (up to the
16 MiB body cap) and dispatch then repeated that work, a 2x
allocation amplification available to scoped principals even for
denied operations.

facade_resolved_tool now reads at most the leading command-path
tokens (6 tokens, 64 bytes each - bounds the registry test proves
cover every path and alias rewrite), so the stream decision costs
O(1) regardless of body size; dispatch keeps the one exact full
resolution. The divergence class is harmless by construction: the
resolver chooses stream framing, never authorization, and every
out-of-bound shape refuses to plain JSON rather than resolving
differently - pinned by a registry-wide equivalence test (every
spelling resolves identically on its own lane and refuses the other
two) plus near-cap-tail, oversized-token, and non-string-token pins.

### DIFF
--- a/src/bin/caller/mcp/facade.rs
+++ b/src/bin/caller/mcp/facade.rs
@@ -1377,18 +1377,39 @@ pub(crate) fn plan_for_meta(meta: &str, args: &serde_json::Value) -> Result<Plan
     })
 }
 
+/// Bounds for the prefix-only stream-mode resolution below: enough
+/// leading tokens to cover every registered command path and every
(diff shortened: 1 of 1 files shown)


lovon-spec:
@codex review

PR #903, opened 2026-08-31
[Superseded by #904] mcp: serve the effective skill catalog over Skills Over MCP

lovon-spec:
## Work in progress — do not queue

Implements the draft `io.modelcontextprotocol/skills` extension on Intendant's authenticated HTTP MCP endpoint.

Current direction:

- advertise the extension during `initialize`
- serve the complete effective catalog through `skills/list` and `skills/get`
- serve every declared `SKILL.md` and support file through `resources/read`, with SHA-256 digests and byte sizes
- derive the catalog from enabled built-ins, enabled-and-ready plugin payloads, and verified owner-added user skills
- keep the canonical Intendant catalog unbounded by OpenAI's current five-named-skill importer limit
- provide an explicit OpenAI projection using up to five imported packages while preserving every source skill's relative resource tree and refusing rather than truncating

The connector cannot currently convert this PR back to draft. Treat this body as the draft marker. Auto-merge remains disabled until the full test and Codex review loop is complete.

## Validation in progress

- targeted Skills Over MCP unit tests
- `cargo test -p intendant --bins`
- library crate tests
- `cargo test -p intendant --test e2e`
- `cargo clippy --workspace -- -D warnings`
- live MCP dogfood against the branch binary
- fresh Codex code and security review loop before merge-queue entry

PR #904, opened 2026-08-31
mcp: serve the effective skill catalog over Skills Over MCP

lovon-spec:
## What

- advertises `io.modelcontextprotocol/skills` from the HTTP `/mcp` initialize response
- serves paginated `skills/list`, `skills/get`, and manifest-backed `resources/read`
- derives the catalog from the daemon's effective skill set: enabled builtins, enabled/ready plugin payloads, and verified owner-added user skills
- preserves complete parsed frontmatter, safe resource paths, and SHA-256 digests
- adds an explicit `skill_profile=openai` projection that folds the complete logical catalog into one imported skill package rather than inheriting OpenAI's current five-named-skill ceiling
- preserves the existing MCP tools transport behind a thin extension wrapper

## Review fixes carried from #903

- the 3,077-line gate extraction is now an isolated, verbatim pure-move commit (`d7f611d`)
- the initialize response no longer advertises the standard `resources` capability without `resources/list`
- aggregate `skill://intendant/intendant-skills/...` URIs self-select the OpenAI projection, so follow-up `skills/get` and `resources/read` calls do not need to repeat `profile=openai`

## Trust and lifecycle

Skill reads use the authenticated MCP principal and the existing `stats.read` operation gate. Disabled builtins/user skills, inactive plugin skills, stale user-library records, and name collisions do not enter the effective catalog. The OpenAI projection refuses rather than silently truncating if the resulting package exceeds the importer’s hard byte/file ceilings.

## Validation

Hermetic unit coverage pins:

- a canonical catalog with more than five entries and a verified owner-added skill
- list/get/resource parity and digest verification
- the OpenAI aggregate containing every effective builtin and user skill body
- profile-free follow-up reads for aggregate URIs
- unsafe paths and malformed cursors
- initialization extension advertisement derived from the implemented method set

Supersedes #903, which was used to validate the initial implementation before the history was reconstructed to satisfy the repository's pure-move rule.


lovon-spec:
@codex address that feedback and update this PR branch. Preserve the existing three-commit history, especially the first pure-move commit; add new commits only.

Please implement all three review findings:

1. Keep `skills/list` and `skills/get` metadata reads at the existing `StatsRead` floor, but authorize `resources/read` separately at `FilesystemRead` (or the repository’s exact equivalent operation for reading owner-authored bytes). Do not let a stats-only principal receive any `SKILL.md` or support-file content.
2. Move every synthetic OpenAI projection URI to a collision-free authority, e.g. `skill://intendant-openai/<package-name>/...`, and make profile inference accept only package names `intendant-skills` and `intendant-skills-2` through `-5`, with only `SKILL.md` or `references/...` paths. Canonical owner skills must remain under `skill://intendant/<name>/...`, including an owner skill literally named `intendant-skills`.
3. An incompatible legacy user skill must never fail the whole effective catalog. Isolate and warn/report that one user entry while continuing to serve unrelated builtins/plugins/users. Keep shipped builtin/plugin catalog errors fail-loud, and add a hermetic regression test with one incompatible user skill plus a valid builtin/user entry.

Also complete the implementation against the current `modelcontextprotocol/modelcontextprotocol` `sep/skills-extension` branch and the current OpenAI MCP skill importer rules. The older development branch `agent/skills-over-mcp` at `ab7fffb1b78bdce6d086c52ce6d33112b71cc4f2` contains useful near-complete code to port, but do not copy its history or temporary workflow. Required final behavior:

- `skills/list` and `skills/get` return `resultType: "complete"`.
- Every listed resource has `uri`, SHA-256 `digest`, and raw-byte `size`; the fetched bytes must match both.
- Preserve complete frontmatter and ensure the JSON frontmatter is field-for-field identical to the fetched canonical `SKILL.md`, including quoted scalar decoding/re-emission.
- Enforce the SEP per-skill limits of 512 resources and 16 MiB without truncation.
- Keep the canonical catalog unbounded in skill count and paginated.
- For `skill_profile=openai`, return at most five uniquely named, tree-preserving packages. Always make the first `intendant-skills` package an any-task router carrying a complete cross-package catalog; place each source skill atomically under `references/skills/<name>/` with its original relative support-file tree intact so relative references still work.
- Enforce OpenAI’s 100-file/package, 256 KiB root `SKILL.md`, 1 MiB/support-file, 5 MiB/package, and 8 MiB combined generated-archive budget (including a conservative ZIP-overhead estimate). Refuse explicitly rather than silently omitting/truncating.
- Keep tests in inline `#[cfg(test)]` modules per AGENTS.md; remove the separate `skills_over_mcp/tests.rs` module/file if porting it.
- Update initialize instructions to describe at most five tree-preserving imported packages; continue omitting the base `resources` capability because `resources/list` is not implemented.
- Do not add or modify GitHub Actions workflows.

Run `cargo fmt --all`, targeted package tests for Skills Over MCP and the HTTP gate, `cargo check -p intendant --bins --locked`, and Clippy if practical. Commit the result to this PR branch and report the commands/results.

PR #910, opened 2026-09-01
Export bounded computer-use transcript evidence

lovon-spec:
## Summary

- export the bounded task’s redacted transcript events in its signed receipt
- validate event ordering, kinds, size bounds, and transcript digest linkage
- independently bind attestation-stage transcripts to the prior stage digest

The exported events contain only hashes, counts, action kinds, display identifiers, and status labels; they do not contain typed text, pixels, cookies, wallet data, or raw model output.

## Verification

- `cargo test --bin intendant --locked --quiet` (6,104 passed; 10 ignored)
- `cargo clippy --bin intendant -- -D warnings`
- focused bounded-task tests (11 passed)
- `git diff --check`

PR #908, opened 2026-09-01
Add bounded computer-use proof task receipts

lovon-spec:
Adds an owner-only `intendant ctl cu task` lane for evidence capture on an exact daemon-owned display generation and attempt-leased local CDP workspace.

The stage path exposes only bounded native computer actions; attestation gets one internally captured frame and rejects every later action. No function tools, shell, filesystem, browser API, delegation, or escalation are sent to the provider. Resource bindings are checked around every action, concurrent tasks on one display generation are refused, scratch space is private and removed before return, and the compact receipt binds all fields plus frame/action/task/result lineage.

Also wires the project CU provider config through daemon MCP state, makes virtual-display generations queryable exactly, adds the CLI/facade surface, and documents the contract.

Verification:
- `cargo clippy --bin intendant --no-default-features -- -D warnings`
- `cargo test --bin intendant --locked` (6,089 passed; 10 ignored)
- 11 focused bounded-CU hostile-path tests
- `cargo fmt --all -- --check`

PR #914, opened 2026-09-04
Bind managed browser CDP to its profile endpoint

lovon-spec:
## Summary

- eliminate the reserve-then-release loopback port race in managed CDP browser launches by asking Chromium to allocate its own port
- bind readiness and every later liveness check to the exact `DevToolsActivePort` record written inside the launched browser profile
- validate browser and page WebSocket identities against the exact numeric loopback host, port, browser token, and target ID
- use a proxy-free, redirect-free, timeout-bounded CDP HTTP client and cap JSON responses at 1 MiB
- reject stale, symlinked/reparse, malformed, oversized, changing, or mismatched endpoint state

## Security finding

The merged #908–#910 tree reserved an ephemeral loopback port, dropped the reservation, and then launched Chromium on that port. A concurrent local listener could acquire the released port. Because readiness checked only a responsive `/json/list` while the spawned Chromium process remained alive, an unrelated listener could satisfy startup and later liveness checks without being the launched browser.

This change removes the handoff window and requires the endpoint advertised by the exact fresh profile to agree with `/json/version`, `/json/list`, the stored workspace port and URLs, and the still-live registered child before and after probing.

## Verification

- `cargo fmt --all -- --check`
- `git diff --check`
- `cargo test --locked -p intendant --bin intendant local_cdp_client_is_proxy_free -- --nocapture`
- `cargo test --locked -p intendant --bin intendant browser_workspace::tests:: -- --nocapture` — 25 passed
- `cargo test --locked -p intendant --bin intendant bounded_cu_task::tests:: -- --nocapture` — 13 passed
- `cargo test --locked -p intendant --bin intendant mcp::tools_bounded_cu::tests:: -- --nocapture` — 11 passed
- `cargo check --locked -p intendant --bin intendant`
- `cargo clippy --locked -p intendant --bin intendant -- -D warnings`

The new hostile cases cover explicit proxy bypass, redirect refusal, response-size limits, stale endpoint cleanup, malformed endpoint bytes, symlink rejection, foreign ports, and mismatched browser/page WebSocket identities.

### DIFF
--- a/src/bin/caller/browser_workspace.rs
+++ b/src/bin/caller/browser_workspace.rs
@@ -19,6 +19,9 @@ static GLOBAL_BROWSER_WORKSPACES: OnceLock<SharedBrowserWorkspaceRegistry> = Onc
 
 const CDP_STARTUP_TIMEOUT: Duration = Duration::from_secs(8);
 const CDP_LIVENESS_TIMEOUT: Duration = Duration::from_secs(2);
+const CDP_JSON_MAX_BYTES: usize = 1024 * 1024;
+const DEVTOOLS_ACTIVE_PORT_FILE: &str = "DevToolsActivePort";
+const DEVTOOLS_ACTIVE_PORT_MAX_BYTES: u64 = 4 * 1024;
 const BROWSER_EXECUTABLE_ENV: &str = "INTENDANT_BROWSER_WORKSPACE_EXECUTABLE";
 const LEGACY_BROWSER_EXECUTABLE_ENV: &str = "INTENDANT_BROWSER_EXECUTABLE";
 // macOS-only system-browser escape hatch; other platforms' discovery path never consults it.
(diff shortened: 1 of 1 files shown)


lovon-spec:
Manual follow-up review found and fixed a boundedness defect in the readiness retry path: profile-endpoint churn used an early `continue`, which bypassed both the shared startup deadline check and the retry sleep. Commit 6d84156 moves deadline enforcement to the loop boundary, binds both CDP HTTP requests with `timeout_at`, and adds live local-server regressions for endpoint churn and stalled responses.

Local gate: `cargo fmt --all -- --check`; 27 browser-workspace tests; 13 bounded task tests; 11 bounded MCP tool tests; `cargo check --locked -p intendant --bin intendant`; and `cargo clippy --locked -p intendant --bin intendant -- -D warnings` — all passed.

PR #924, opened 2026-09-14
feat(mcp): add rmcp 3 Tasks for remote commands

lovon-spec:
## Summary

Upgrade rmcp from 1.4 to 3.3.0 and expose MCP Tasks for `remote_command start` on the trusted stdio transport. Tasks adapts the existing remote executor, including its real cancellation and caller ownership checks. The daemon HTTP `/mcp` gateway retains its existing protocol and authorization behavior.

## Protocol and ownership

- A non-cloneable `StdioTaskServer` owns a fresh task store for each connection. Only this wrapper advertises Tasks and adds `remote_command` to the trusted owner stdio router; ordinary tools retain their original routing.
- Negotiate Tasks through `io.modelcontextprotocol/tasks`. Preserve the existing base protocol revisions and newer-client fallback without advertising the SDK's newer subscriptions/listen contract.
- Return a task only for an admitted `remote_command start` from a Tasks-capable client. Legacy clients, other operations, and admission failures retain their existing responses.
- `tasks/get` exposes authoritative job state and the terminal result. Cancellation requests real executor cancellation; completion wins a cancellation race. Remote commands do not request input, and there is no separate `tasks/result` endpoint.

## Lifecycle and compatibility

- Serialize dispatch with cancellation so a queued/preparing job cannot be declared cancelled while its Start frame is being sent.
- Track cleanup independently from the SDK result future. TTL expiry and disconnect request real cancellation, continue observing the job, and retain capacity until cleanup finishes.
- Bound each connection to 32 active, retained, or cleaning-up tasks, with a three-hour TTL and one-second suggested polling. Document opportunistic sweeping and in-memory retention limits.
- Migrate the SDK's content/resource types, non-exhaustive constructors, and resource response types. Declare the tagged remote-command schema's object root before SDK validation. Preserve outbound MCP credential filtering and untrusted result/error envelopes.
- Document that real Cloud execution requires the stdio instance's own authenticated gateway and existing environment, enrollment, and TLS configuration; workers attached to another daemon are not imported.

## Regression coverage and CI reliability

- Keep the 11 SDK Tasks/protocol tests and four executor race tests in inline `#[cfg(test)]` modules, as required by the repository. Cover negotiation, legacy fallback, lifecycle races, retention, shutdown, and connection/session isolation.
- Fix the Linux external-CU acceptance guard: sibling CI jobs may start or finish while this test runs. Preserve full identities for runner listeners and the owning job worker, retain the whole-host guard outside CI, and save before/after runner evidence. Seven hermetic regressions cover worker turnover, PID reuse, listener changes, missing ownership, and process ancestry.
- Isolate the existing Git-tag and macOS sandbox-consent fixtures from host signing settings and temporary-path grants without changing user configuration or sandbox policy.

## Validation

- Local focused Python suite: **12 passed**; Rust formatting and whitespace checks passed.
- CI for `8a61a3f6`: **all nine PR checks passed**. [Cross-platform run](https://github.com/intendant-dev/Intendant/actions/runs/34832965549) passed the full Linux binary/library/E2E suites, Clippy, smokes, browser isolation, and external-CU acceptance; macOS and Windows build checks passed. WASM drift, repository integrity, generated HTML, and audit checks also passed.
- The original implementation also passed the full macOS binary, library, E2E, and Clippy battery before this follow-up. No real Cloud job was validated.

Auto-merge remains disabled.


lovon-spec:
@codex review

PR #925, opened 2026-09-14
feat: opt-in macOS background window computer use

lovon-spec:
WIP: process-generation-bound windows, process-local CGEvents, AX grounding, window-only capture, MCP facade. No foreground/HID/clipboard fallback. Integration and hermetic checks in progress. Native app compatibility is not verified; do not merge/deploy until tests pass.


lovon-spec:
## Native acceptance follow-up: keep this PR draft

Tested head `17e316c13d06a7db3ca299039480e7d42757811c` on the plugin Mac using a new disposable AppKit accessory application, not the user's documents or the installed Codex app. No daemon was installed or restarted, and no permissions or user settings were changed.

The existing native component harness passed **8 tests** (the live-frontmost test stayed excluded). I then built a small executable from those same production `background_cu`, native dispatch, AX and process-generation sources and targeted only the fixture PID/window. This exercises native input and AX, **not** the HTTP facade or production screenshot capture. The component harness still stubs capture and the key-name parser; neither was used in these input tests.

| Case | Observed result |
|---|---|
| Exact-window AX read | Passed; only the selected fixture window's tree returned, with window-local coordinates. |
| Background text entry | Passed; the fixture text field actually became `fixture-text-924`. |
| Semantic button click | Passed; AXPress incremented the fixture button counter from 0 to 1. |
| Raw click into an AppKit canvas view | **No application effect:** dispatch reported `injected` as documented, but the fixture's mouse-down counter remained 0. |
| Stale process-generation selector | Correctly refused before input. |

The foreground application PID, pointer coordinates and clipboard change counter were unchanged across all five before/after samples. The fixture remained inactive and was terminated/reaped by the harness. These are sampled observations, not a claim of an OS-wide focus lock or continuous non-interference.

**Conclusion:** this confirms real value in the AX/background typing path, but does not establish generic background mouse compatibility. Keep #925 draft; do not equate green CI or `injected` with a successful canvas action. The raw-canvas path needs investigation before broader promotion. Chromium, multi-window/focus races, Retina/occlusion, production ScreenCaptureKit capture and end-to-end HTTP/plugin acceptance remain open.

This does not make the separate CGVirtualDisplay work a replacement: monitor creation alone does not supply independent input focus or sandbox the user's session.

Local reproduction/evidence on this Mac: `/tmp/intendant-bg-accept.sUKrZF/Fixture.m`, `/tmp/run-bg-accept.py`, `/tmp/bg-accept-main.rs`, `/tmp/bg-native-acceptance.json`, and `/tmp/background-cu-component.log`. The review worktree is `/tmp/intendant-background-review`; its tracked sources were not changed.

PR #930, opened 2026-09-16
feat(macos): daemon-owned monitors with generation-bound read-only capture

lovon-spec:
## Summary

Follow-up to #927, stacked on its refreshed lifecycle branch. #929 is merged through the protected queue; this branch contains that security fix. The user-reviewed #928 is unchanged and remains independent.

This implements two connected slices: a daemon-owned main-thread helper/broker, then generation-bound read-only SCK capture through existing MCP tools. `create_virtual_display`, `take_screenshot`, and `destroy_virtual_display` work over the existing HTTP router with original IAM plus shared-session display authority.

## Boundaries

- One lazy private-pipe helper; bounded queue of eight, two monitors maximum, even dimensions 64..4096. Native objects stay on the helper main thread.
- Public logical IDs are reserved separately from private handles/native IDs. Exact opaque generations are required. Unsupported/raw-ID aliases are refused before input, AX, shared views, readiness, browser placement or forwarding.
- Capture preflights existing Screen Recording permission, selects only the owned native ID, disables input and cursor overlay, validates frame dimensions, and publishes complete non-clobbering 0600 PNGs.
- Cancellation retains native ownership through stop and cleanup. Failed helper/uncertain teardown permanently retires the broker; no automatic respawn or primary-display fallback.
- Receipt commit means local tool-result construction, not a network acknowledgment. Lost committed handles can occupy capacity until broker shutdown; recovery/status is a separate follow-up. `list_displays` is unchanged.
- No keyboard/mouse, independent focus/clipboard, AX, app placement or streaming is claimed. A virtual monitor shares WindowServer and is not a security sandbox. Existing Linux Xvfb behavior is unchanged.

## Native acceptance

Passed on the plugin Mac (macOS 26.4.1, build 25E253, arm64) using a separate temporary-HOME HTTP daemon, not the installed service. The disposable fixture refuses the primary display. The 800x600 screenshot returned exact red/green/blue/white quadrant samples; authentication, 0600 artifacts, input refusal with wait-only actions, stale generation refusal, explicit teardown, parent EOF cleanup and restoration of the original display inventory passed. No existing app/document was used as a target.

Reproduction: `scripts/verify-macos-monitor-http.py` with `tests/fixtures/macos-monitor/pattern.m`. The simpler private-entry smoke is `scripts/macos-monitor-smoke.py`. Both require explicit native opt-in and are excluded from default tests.

## Validation

[...]

PR #927, opened 2026-09-15
feat(macos): add guarded CGVirtualDisplay lifecycle primitive

lovon-spec:
## Scope: first CGVirtualDisplay slice, not a virtual-desktop product yet

Independent branch from main `05f9674`, which includes merged #924. This PR deliberately does **not** include HTTP Tasks or #925's existing-window background CU.

Add a platform-layer, runtime-probed macOS `CGVirtualDisplay` lifecycle primitive and an explicitly opt-in native smoke example. Leave `create_virtual_display`, daemon/dashboard capability reporting, capture routing, and CGEvent input unchanged.

A virtual monitor shares the logged-in WindowServer session, focus, cursor, clipboard and application state. It is not a security sandbox, a separate login, or an implicit user-display grant. Hotplug can rearrange windows or persist OS monitor preferences.

## Implementation and safety

- Runtime class/selector/full ABI validation before private native calls. No static private-class linking or fallback to a physical display. Missing/changed ABI fails closed.
- Main-thread, non-Send/non-Sync owner; one owner per process; at most two monitors; dimensions 64–4096; single 1x/60 Hz mode.
- Opaque owner-bound, non-wrapping generations. Native display IDs are observations of retained objects, not permission tokens or lifecycle lookup keys.
- Small Objective-C ARC/exception boundary, Rust RAII ownership, partial-creation cleanup, explicit removal and owner-drop teardown. Unconfirmed native cleanup disables future creation for the process lifetime.
- Bounded fresh online-display enumeration for readiness/removal. Failed or potentially truncated inventory cannot establish removal. Native OS calls themselves are not interruptible.
- Default tests use fake native objects and bounded public NSObject metadata, never native monitor creation, GUI input or TCC changes. The manual example requires the exact `--create-shared-session-monitor` flag.

## Native acceptance

On the plugin Mac, macOS 26.4.1 (25E253), arm64:

1. The initial smoke safely rejected a wrong ABI assumption: NSObject `release` is `oneway void` (`Vv`), not plain `v`. Fixed the exact signature and added a regression; no broad signature-check bypass.
2. A monitor was created, but cached `CGDisplayIsOnline` flags did not confirm its removal. Replacing that observation with fresh `CGGetOnlineDisplayList` made explicit removal observable (about 84 ms in the successful run).
3. The final native smoke created 1024×768, observed explicit removal, created a replacement, rejected the stale first-generation handle, and observed RAII removal. The original primary display ID and display inventory matched before/after.

[...]

PR #931, opened 2026-09-16
feat(macos): recover owned monitor handles and report read-only readiness

lovon-spec:
## Scope

Stacked on #930 (`feat/macos-monitor-controller`), which is stacked on #927. This PR adds two read-only slices; it does not add input, placement, streaming, deployment or another permission grant.

- `inspect display monitors` / typed `list_macos_monitors`: owner-surface-only recovery of committed logical generation handles. Existing `list_displays` is unchanged. Empty inspection does not start an actor/helper. Scoped user-display authority is not sufficient to enumerate daemon-wide recovery handles.
- Exact `display_readiness`: uses the existing DisplayView and shared-session authority checks. Resolves the owned generation without capturing. Preserves target/summary/five-layer readiness shape; capture remains unverified, input and streaming unsupported, overall readiness false.

The bounded actor serializes inspection behind receipt commit/rollback and cleanup. Stale/foreign handles do not select another display. Failed verification retires the existing helper without respawn; busy/retired snapshots contain no usable handles. Public descriptions exclude native/helper IDs, PIDs and paths.

## Validation

Final local macOS battery passed with compile governor intact: **6,466 binary tests** (10 existing ignores), **1,020 library/acceptance tests** (3 existing ignores), **55 E2E tests**, workspace Clippy `-D warnings`, formatting and whitespace checks. An initial new test incorrectly expected an empty properties map in the unchanged legacy schema; the assertion was corrected to its actual omitted-properties shape and the full binary suite rerun successfully.

The opt-in temporary-HOME HTTP/native fixture passed on 2026-09-16 with `--check-recovery`: empty initial inventory, discarded/recovered creation handle, exact 800x600 RGB/white pattern capture, status unverified before/after capture, 0600 image artifact, removal from inventory on destroy, stale refusal, and parent-EOF cleanup restoring the initial native inventory. The client discards its handle; this is not an actual tunnel-disconnect test. No user app documents were used, no input was injected, no permissions were changed, and the installed daemon was not replaced.

The full cross-platform workflow filters PRs targeting main. It does not run on this stacked base; do not treat a green integrity check as the full matrix. Retarget/revalidate after prerequisite merges. Cloud acquisition remains unconfigured on this machine; no Cloud execution was used.

## Next boundary

Explicit authorized app/window binding and placement on an owned monitor, then separately validated targeted input with real native/Chromium effect and human-interference fixtures. #925 remains separate; a virtual monitor does not by itself fix background canvas clicks or provide focus/clipboard isolation.


lovon-spec:
Landing preparation: independent read-only review of the combined monitor stack at 222a68f found no concrete merge blockers. The prior full local and native fixture validations passed. Retargeted to main; reopening immediately to trigger the required pull_request CI matrix, which did not run for the old stacked base. No code/history changes and no check bypass.

PR #932, opened 2026-09-16
feat(macos): bind and place explicit windows on owned monitors

lovon-spec:
## Summary

Completes the explicit owner-only window binding/placement slice on the owned CGVirtualDisplay helper. list_macos_windows lists only an explicitly selected PID; bind_macos_window requires the full identity plus a one-use retained candidate token; place_macos_window verifies a fully contained rectangle using AX and CG readback; unbind_macos_window releases the exact reference.

## Review fixes

Independent review found four defects, all corrected with regressions: receipt expiry/send arbitration; listed-window ID reuse before binding; loss of the latest partial geometry observation; and missing effects-unconfirmed evidence on late failures. New binding/candidate selectors also fail closed in generic display routes rather than falling through to legacy defaults.

Readback settling is observation-only, bounded to 250 ms / 20 polls within the original operation budget. The position and size setters are each attempted once; no action retries or focus restoration.

## Acceptance and validation

Native HTTP acceptance passed on macOS 26.4.1 arm64 using a separate temporary-HOME daemon and disposable panel: refreshed/consumed/replaced candidates rejected, exact placement verified, independent CG readback matched, stale binding refused, fixture reaped, original monitor inventory restored. No existing app/document was targeted.

Final local: 6,499 binary tests, 1,020 required library/acceptance tests, 55 E2E tests; 63 focused monitor tests; workspace Clippy -D warnings, formatting and whitespace passed. Existing ignores: 10 binary and 3 library. Compile governor untouched.

## Boundaries

No keyboard/mouse input, clipboard, activation, independent seat, streaming or general application compatibility is claimed. Shared WindowServer authority stays owner-only for window operations; scoped display grants do not permit them. The installed daemon is unchanged.

Reproduction and retained failure analysis: docs/design/macos-monitor-window-placement.md. Prerequisites #927, #930, #931 and HTTP Tasks #928 are merged. New-head CI must validate this PR against main.


lovon-spec:
The merge queue ejected #932 after Windows E2E eight_fire_wave_staggers_and_daemon_stays_responsive failed at the existing minted-item-id assertion (run 35075533303). macOS and Linux merge-group jobs passed. Retaining the failure and rerunning that failed Windows job once without altering assertions; root cause is not established. New controls work remains separate #933.

PR #935, opened 2026-09-16
fix: keep dogfood reporting developer-only and opt-in

lovon-spec:
## Why

Correct the distribution/policy boundary introduced by #926. Dogfooding is an explicitly opted-in developer facility, like the pre-existing internal skill system—not ambient end-user behavior. Keep the structured local report infrastructure, dedupe, narrow operation, and trusted provenance.

## Default-off behavior

- Remove `intendant-dogfood` from `skills/`, `BUILTIN_SKILLS`, and the default ChatGPT plugin template/manifest. Facade `docs` no longer serves these instructions.
- Capture an explicit daemon-process opt-in, **`INTENDANT_DEV_DOGFOOD=1`**, when the MCP server starts. Missing values and every other value stay off; changing it requires restart.
- Without that opt-in, omit `report` from all HTTP MCP profile listings and reject cached/direct report calls **before any Agenda write**, including owner/root callers. Request parameters, a plugin package, or IAM authority cannot enable the feature.
- Restore the default facade to six tools; `report` is an optional seventh on opted-in developer daemons.
- Remove `feedback.write` from the ordinary builtin `role:operator` and the dashboard fallback. Existing builtin roles normalize to the corrected definition; custom owner-defined roles are preserved.
- Keep root/agent-operator's existing authority ceilings, but neither bypasses the independent feature opt-in. Scoped developer callers still require explicit applicable `feedback.write` authority; the feature flag grants none.

## Developer workflow

The only instruction layer is `skills-internal/skill-dogfood-feedback`, installed deliberately through the existing dev-skills installer. It now covers skills, facade/plugin/relay/daemon issues and concrete efficiency recommendations. It explicitly stops silently if reporting is absent/disabled/denied; it must never self-enable, widen permissions, or fall back to generic Agenda writes.

For an explicitly developer-only ChatGPT package, `configure_plugin.py --dev-dogfood` includes that internal skill. The default generator includes no dogfood skill. Packaging does not enable the daemon or grant IAM authority.

## Upgrade/opt-out

The existing ownership-aware skill reconciler sweeps retired **daemon-marked** builtin copies from both global roots. User-owned directories/symlinks and existing feedback records are not deleted. Rebuild/reinstall old ChatGPT packages without `--dev-dogfood`; restart the updated daemon without `INTENDANT_DEV_DOGFOOD=1` to refuse even cached clients. Documentation covers these independent controls.

## Validation

Completed on GitHub's runner:
- `cargo fmt --all`
- `cargo run -p app-html-assembler --locked` (generated `static/app.html` included)
- `python3 -m unittest scripts/test_configure_intendant_plugin.py scripts/test_intendant_mcp_relay.py` — 10 passed
- `bash -n scripts/install-dev-skills.sh`
- `git diff --check`

Added Rust regressions for exact opt-in parsing, default-off listings across profiles even after enabled schema caching, disabled owner/scoped and repeated calls making no ledger writes, enabled create/merge behavior, stale builtin permission migration, default skill/docs exclusion, and ownership-safe upgrade cleanup. The standard PR Rust/Clippy/cross-platform checks are still to run; no claim that those passed yet.

The six new Python packaging tests also run in the existing required repo-integrity job. Temporary preparation workflow/patch files have been removed from the final diff; no main history was rewritten.


lovon-spec:
Fixed the initial CI compile error in a new regression test: the dogfood docs assertions now live inside the facade's existing test module, without widening production module visibility. The final diff contains no temporary preparation files, and CI is rerunning on `b51cbb7f7b890871fbbe1332ecd31c54a8988bd0`.

@codex review

@codex security review

PR #933, opened 2026-09-16
feat(macos): inspect and act on exact bound-window controls

lovon-spec:
## Summary

Two owner-only slices on merged #932: bounded exact bound-window element inspection, and one-use AXPress / text-value actions. Existing DisplayView/DisplayInput IAM and real caller trust are preserved. No global input, clipboard, activation, permission changes or daemon deployment.

The helper retains exact control/ancestor objects, rechecks labels, geometry, security and current membership, and consumes tokens before dispatch. Text success requires exact readback; a press is only dispatched/effect-unconfirmed. Strict parsing, queue/dispatch cancellation and late-result evidence are regression-covered.

## Validation

Final local: 6,547 binary tests (10 existing ignores), 1,022 required library/acceptance tests (3 existing ignores), 57 E2E tests; workspace Clippy -D warnings, formatting and whitespace passed. Compile governor unchanged. A test-only import error was fixed and the complete binary suite rerun.

Native semantic run passed on an isolated temporary-HOME daemon: independent Unicode text readback, button counter exactly once, stale/replaced/replay/unbind refusal, secure/disabled omission, monitor restoration and fixture cleanup. Later final-build runs refused on listing/focus observation; those failures are retained, not counted as passes. The successful native run preceded the final optional-label error tightening. General AppKit/Chromium compatibility is NOT established; missing optional subroles deliberately refuse.

## Review corrections

Addressed the saved review findings on dispatch/cancellation ordering, changed labels, reciprocal child membership and normal leaf absence. This continuation also fixed strict empty press parsing, ignored AppKit fixture wrappers, mutation IPC budgeting (max 500ms inside 4s), optional-label error classification and fixture cleanup ordering. No action retries or weakened identity/focus guards.

Fresh local automated review was quota-blocked; no new independent review result is claimed. Please review this exact published head; source contract and reproduction are in docs/design/macos-bound-window-actions.md. Installed daemon remains unchanged.

PR #936, opened 2026-09-18
feat(macos): support ordinary AppKit control metadata safely

lovon-spec:
# Standard AppKit control compatibility

Follow-up to #933. Replace fixture-only assumptions about optional AX subroles with documented optional-attribute decoding: distinguish absence from failed IPC or malformed values. Continue excluding explicit secure/password roles and subroles, and additionally exclude AXContainsProtectedContent subtrees before labels, children or values. Exact retained identity, membership, current permissions, geometry, focus and single-use tokens remain required.

Add native acceptance with unmodified NSTextField, NSButton, NSSecureTextField and ordinary AppKit hierarchy, without synthetic subroles or parent/children overrides. Record baseline failures before changing the adapter. No raw input, activation, clipboard, permission changes or live-daemon installation.

Native acceptance passed for both profiles below; repository validation and review status are recorded with the PR.

## Reproduced baseline

On the unchanged #933 head e2354b2, a fixture compiled with
`-DINTENDANT_STANDARD_APPKIT` refused `required AX attribute AXSubrole unavailable`.
The standard profile uses native NSTextField, NSButton, NSSecureTextField and
NSView classes, plus a nonactivating panel whose only overrides prevent becoming
key/main. No AX subrole, parent or children override is used in this profile.

## Contract and sources

The installed Apple SDK's AXAttributeConstants.h documents AXSubrole as required
only when the role alone is insufficient. NSAccessibilityProtocols.h declares
accessibilitySubrole nullable and accessibilityProtectedContent as the Boolean
for NSAccessibilityContainsProtectedContentAttribute. The native adapter admits
only documented absent values, not errors; explicit secure/password metadata or
true AXContainsProtectedContent excludes the entire subtree. This is reported
application metadata, not an isolation boundary or a secret detector.

Apple reference: https://developer.apple.com/documentation/appkit/nsaccessibilityprotocol/accessibilitysubrole()
Apple reference: https://developer.apple.com/documentation/appkit/nsaccessibilityprotocol/isaccessibilityprotectedcontent()
Apple hierarchy reference: https://developer.apple.com/library/archive/documentation/Accessibility/Conceptual/AccessibilityMacOSX/OSXAXmodel.html

The unchanged control engine checks safety again before and after action dispatch
and before reading text back. The native fixture includes secure and disabled
controls and an explicitly protected container. Existing strict limits, owner/IAM,
retained membership and no-retry/focus rules remain unchanged. Native results are recorded below.

## Native validation, 2026-09-18

The standard AppKit profile passed against the new adapter on macOS 26.4.1 arm64. The normal field reported no subrole. Exact Unicode text, independently observed one-time button action, refresh/replay/replacement/unbind refusal, monitor capture/recovery and cleanup passed.

Protected-content validation made a synthetic subtree visible while explicitly unprotected, retained its element token, re-protected the container, and verified that the old token was refused before any action. A fresh inventory again excluded the whole subtree.

The original explicit-subrole controls profile also passed. A separate combined run stopped in the older placement fixture with an application-window availability refusal; that run remains a failure, not a controls pass. All runs restored display inventory and cleaned up their fixtures.

```sh
clang -DINTENDANT_STANDARD_APPKIT -fobjc-arc -framework AppKit -framework CoreGraphics tests/fixtures/macos-monitor/controls.m -o /tmp/intendant-standard-appkit-fixture
python3 scripts/verify-macos-monitor-http.py --bin /absolute/built/intendant --fixture /tmp/intendant-monitor-pattern --controls-fixture /tmp/intendant-standard-appkit-fixture --report /tmp/standard-appkit-proof.json --allow-shared-session-monitor --check-recovery
```

Without the compilation define, the same source retains the original explicit-metadata fixture profile. It uses the same acceptance harness. Neither profile establishes arbitrary app/Chromium/canvas compatibility or independent focus/clipboard isolation.

## Local regression validation

Passed: 6,550 binary tests (10 existing ignores), 1,022 required library/acceptance tests (3 existing ignores), 57 E2E tests, workspace Clippy with warnings denied, formatting and whitespace. All compiler-governor settings were left intact. The AX adapter subset passed eight tests with one existing GUI-only ignore. No ignored native test was counted as executed.


lovon-spec:
Published head 0a3334f was rebuilt and passed standard-AppKit native acceptance, including protected-subtree revalidation, exact text/button effects, stale tokens, capture and cleanup. Linux CI failed in unchanged handover::tests::successor_port_never_names_a_draining_or_dead_target at b.poll_acquire(0); the handover tree is identical to merged main/#933 and no macOS production code is built on Linux. The root cause is not established. Retrying only the failed Linux job once, with no code/test-policy changes. Original failure log retained; not treating it as a pass. Local full battery remains green.

lovon-spec:
The one targeted Linux retry completed successfully with no source changes. All PR checks are now green; GitHub confirms #936 is queued at position 1 on head 0a3334f. Published-head standard-AppKit native acceptance passed, including protected-subtree stale-token refusal, and all fixtures were cleaned up. Original handover-test failure and native negative runs remain retained. No merge bypass or installed-daemon update was performed.

lovon-spec:
Queue ejection investigated: run 35343852948 failed the unchanged handover::tests::dropped_runtime_releases_lease_and_liveness assertion on macbook-b. The prior Linux failure was the companion scheduler takeover assertion. Separate fix #937 reproduces close-only release retaining locks through duplicated/inherited descriptors: both deterministic production-guard regressions failed before the patch; the patched 95-test handover suite passes. Full validation is running. No runner/cache/concurrency settings or test assertions were weakened. Leaving #936 unchanged and unqueued until the independent fix lands; the original CI descriptor interleaving was not traced.

PR #938, opened 2026-09-18
fix(macos): avoid redundant placement writes and admit bounded deep AX trees

lovon-spec:
## Scope

Two bounded fixes discovered during Chromium native acceptance on merged #936: avoid redundant window-geometry writes, and permit depth-16 AX trees without widening node/child/control/wire/time limits. This is NOT a claim of general Chromium semantic input or canvas support.

## Placement correction

A disposable 720x530 Chromium window moved successfully, then rejected an unnecessary AXSize write. Placement now skips a component only when both AX and CG observations already match it exactly, while retaining all pre/post identity, monitor, containment, permission and focus checks. Actual setters still independently check mutability. Results report 0/1/2 attempts, and late receipt failures preserve known zero-write evidence. A deterministic regression failed on the original implementation.

## Bounded traversal and fixtures

The read-only browser diagnostic measured 51 nodes, depth 13 and at most six children. Depth rises from 8 to 16; 128 nodes, 32 children, 16 controls, 16 KiB wire and four-second budget remain unchanged. Tests pin depth 13/16 success, depth 17 refusal and newly protected ancestry refusal.

The opt-in fixture uses an explicitly supplied Chrome for Testing bundle, a new private profile and synthetic CSP-restricted page. Native ownership/cleanup stays with the returned NSRunningApplication. CDP only creates the background test window and independently observes fixture state; tested actions use real ctl/HTTP tools. Placement-only and full semantic profiles are labelled separately.

## Native evidence and remaining boundary

Placement-only acceptance passed: exact one-write placement, independent browser bounds agreement, verified zero-write repeated placement, unbind, browser/profile cleanup, outer monitor capture/recovery and display inventory restoration. The exact published head f878e69 was rebuilt. Its first fresh-fixture run refused unavailable focus before writing; the second fresh-fixture run passed this placement-only profile with one write then zero writes, no sampled browser activation, and confirmed complete cleanup. Both reports remain retained.

The full semantic profile still refuses `retained AX parent changed at depth 7 for AXWebArea`. No Chromium semantic action was dispatched. The ancestry guard is not weakened. App-mode startup activated Chromium despite the nonactivating request; that failed attempt is retained. The replacement background-window profile refuses observed activation. Focus-unavailable and late AppKit metadata-read failures are retained separately, not counted as positive acceptance.

## Validation

Local: 6,562 binary tests (10 existing ignores), 1,022 library/acceptance tests (3 existing ignores), 57 E2E, workspace Clippy -D warnings, formatting and whitespace passed. Six bounded-protocol/subprocess Python harness tests pass. Native fixture compiles.

No live-daemon deployment, user-document/profile access, OS permission change, global input fallback, CI policy change or compiler-governor override. Design, reproduction and limitations: docs/design/macos-chromium-controls.md.


lovon-spec:
Published head f878e6953d61242da392dc2165b2075dcf8a2787 passed the explicitly scoped placement-only native profile: one-write move, matching independent browser bounds, zero-write repeat, exact unbind, browser termination, supervisor reaping, profile removal and original display inventory restoration. Browser was never observed foreground in this run. The first published-head trial refused unavailable focus before input and is retained separately. Full semantic controls remain unsupported at the exact AXWebArea ancestry mismatch; no such pass is claimed. Local required battery and six Python harness tests pass. Submitting only this documented narrow fix scope through the normal protected queue; no installed-daemon or CI-policy change.

PR #939, opened 2026-09-18
fix(macos): retain exact parent ancestry for embedded AX controls

lovon-spec:
## Summary

Follow-up to merged #938: retain the exact omitted AXScrollArea between an AXGroup and its exposed AXWebArea. Both intermediate and originally exposed child are retained and rechecked. The native AXParent/PID/AXWindow check remains exact; no replacement lookup or global input fallback is added.

Targeted membership rechecks avoid re-reading unrelated sibling metadata while preserving raw child membership and all projected-edge witnesses. Existing traversal/IPC/time/wire limits, owner-only IAM, generation-bound handles, one-use tokens and partial-effect reporting remain unchanged.

## Native acceptance

Full Chrome for Testing 153.0.8010.52 semantic acceptance passed on macOS 26.4.1 arm64 through ctl/HTTP/AX. Exact Unicode text, one independent button-counter increment, refresh/replay/control-replacement/same-window-navigation refusals, capture and cleanup passed. Standard AppKit (including protected subtree revalidation) also passed. The successful Chromium run used one discovery call. Fresh private profiles and synthetic fixtures only; no user apps/documents were targeted.

The harness now records bounded read-only startup readiness and independent synthetic readback even when an action reports partial. No mutation retry. Foreground/pointer/clipboard-counter observations were unchanged in the successful Chromium run; this is not continuous isolation.

## Validation

Final local: 6,569 binary tests (10 existing ignores), 1,022 required library/acceptance tests (3 existing ignores), 57 E2E; 96 focused monitor tests; 8 Python harness tests; Clippy -D warnings, formatting and whitespace passed.

Earlier native AX/focus/deadline refusals are retained and not relabeled passes. An initial library doctest resolution failure passed after serial rerun. One unchanged E2E mock-provider test missed expected stderr; full confirmation passed without changing its assertions. Reproduction and limitations: docs/design/macos-ax-parent-ancestry.md.

## Boundary

No raw canvas/keyboard input, activation or clipboard use. No relaxed permissions, runner changes, cache purge, compiler setting override, installation or daemon restart. CI and review for the published head still need to run.


lovon-spec:
Exact published head a46e4b9 was rebuilt and passed the full Chromium 153 semantic-control profile over HTTP: exact Unicode text and independent DOM readback, one button increment, refresh/replay/replaced-element/page-reload refusals, capture and exact cleanup. The browser was never observed foreground. Per-action focus checks passed. Across the whole run, foreground PID and pointer samples changed while clipboard change count stayed fixed; this is not a claim of continuous focus/pointer isolation. The earlier final-source run retained unchanged before/after samples. Standard AppKit regression also passed. All PR checks are now green; protected auto-merge requested for this exact head. No source/CI policy or installed daemon changes since publication.

PR #940, opened 2026-09-18
test(macos): add attributed raw pointer probe and concurrent-owner evidence

lovon-spec:
## Scope

Two test-infrastructure slices after merged #939: a nonposting raw-pointer event-construction probe, and concurrent-owner desktop evidence that does not conflate human activity with agent interference. No production Rust, MCP tools, IAM, input capability, dependency or CI workflow is changed.

## Default nonposting probe

The native fixture creates event objects using a private source, checks the actual private state-table ID and all routing/tag/coordinate fields, then releases them. Thirteen cases pass under `clang -Wall -Wextra -Werror`: negative/positive/fractional coordinates, paired event fields and rejected malformed inputs. It creates no NSApplication/window/display and posts zero events. Native readback initially caught an incorrect private-state-ID assumption; strict JSON checks also caught an integer-vs-Boolean report mismatch. Both were corrected, with negative evidence retained.

The optional live mode requires the exact `--allow-disposable-process-click` flag and can target only `getpid()` and its own disposable panel. It has no arbitrary target arguments, global-post/warp/tap/keyboard/activation fallback or retry. It requires tagged exact down/up receipts and a separate canvas counter, with source/receiver/window/position checks and supervised-child PID binding. The correlation tag is diagnostic, not authority.

**Live mode was compiled but NOT executed because the owner is active on this Mac.** Neither cross-process nor Chromium raw-canvas delivery is claimed. This is not a daemon input implementation.

## Concurrent-owner evidence

The Chromium harness adds a separate assessment of whole-run foreground/pointer/clipboard-counter changes. Their cause stays undetermined; missing data stays unknown; identical samples do not prove continuous isolation. Existing action-level focus guards, foreground-target refusal and pass/fail outcomes are unchanged. Reclassification of the saved #939 report preserves both the desktop changes and its passing action-level focus checks; no GUI replay was needed.

## Tests and boundaries

Sixteen new hermetic Python tests and eight existing browser-harness tests pass. The native default and supervised nonposting runs pass all thirteen cases; extra PID arguments refuse before GUI/input. Output/time limits, private non-overwriting reports, malformed JSON, false-positive receipts, owner movement and unknown observations are covered.

No installed-daemon update, runner restart, permission change or compiler-setting override. Remote compute failed before execution because no Cloud environment is configured; local validation is serial.

Design/reproduction: `docs/design/macos-raw-pointer-probe.md`.

## Final local regression

6,569 binary tests (10 existing ignores), 1,022 required library/acceptance tests (3 existing ignores), 57 E2E tests, workspace Clippy with warnings denied, formatting and whitespace all passed. Both Python suites passed (16 + 8 tests). Native construction-only acceptance passed with 13 cases, no application/window and zero posted events.


lovon-spec:
Published head a30b16a was recompiled for the native construction-only probe and passed all 13 cases with no application/window and zero posted events. The 16 evidence/supervisor tests and 8 existing Chromium harness tests passed again on this head. Full local Rust battery: 6,569 binary, 1,022 required library/acceptance, 57 E2E; Clippy -D warnings, formatting and whitespace passed. Worktree clean.

The owner is active on the Mac. No live click, keyboard input, monitor hotplug or GUI acceptance rerun was performed. The saved #939 report was reclassified without modification: whole-run cursor/foreground changes remain unattributed, and its two native action-level focus checks remain passing. Live self-process delivery and cross-process/Chromium raw-canvas acceptance remain explicitly unverified; this PR enables no production input capability.

